PR #2, opened 2018-07-18
Tidy-up suggestions

eduardo-rodrigues:


henryiii:
Thanks! I've fixed (in theory) the test coverage drop to not "fail" for minor drops in coverage.

PR #11, opened 2018-08-01
Update of links in README.md

eduardo-rodrigues:
There are other links still containing "henryiii" but those are not ready to be updated yet.

BTW, is this file meant to supersede README.rst?


eduardo-rodrigues:
Ah, coveralls also needs to be updated following the transfer henryiii -> scikit-hep ...

PR #16, opened 2019-03-08
Preparing package dependency on particle

eduardo-rodrigues:


eduardo-rodrigues:
There's one test failing but it's in fact the same issue as before this PR was created. I would address this problem in a separate PR, since totally unrelated to the matters at hand here.

eduardo-rodrigues:
This PR will not have all checks green until we get out a new release of ``particle``. To be done asap.

henryiii:
@eduardo-rodrigues , particle no longer allows custom three_charges but pulls them from the PDGID. Unfortunately, this means some of the custom particles loaded here can't be used, since they don't have a valid PDGID code. I think we need to re-add charge as a property/column in the table, and check to make sure they match for the official particles. I'll probably try to improve the behavior as well for repr on a non-valid particle.

We could also make three_charge optional.

eduardo-rodrigues:
Yes, that's a good point. Reminds me of the comment I put in the `convert.py` file in `particle`:
```
    # These items are not very important - can be reconstructed from the PDG ID
    # TODO: maybe first check the consistency between what is read in and what the PDG ID provides (being maniac)?
    del full['Charge'], full['J']
```
Spot on :-).

eduardo-rodrigues:
As agreed, merging this big PR at this stage. The error is rather different from the previous ones, and should be addressed in dedicated developments asap.

PR #20, opened 2019-04-02
Update of Mint .csv file for new specifications in the Particle package

eduardo-rodrigues:
While at it, I also added a README file for the `decaylanguage.data` directory
*and* switched off a test related to work-in-progress code, as it makes little sense to run a broken test until the related code is updated.


eduardo-rodrigues:
@henryiii, any idea how to fix the check in tox.ini? In fact, is it at all needed?

eduardo-rodrigues:
Hi @henryiii, this PR is now a 2-in-1:
- The PR itself.
- Check run on Windows now switched off so that the tests run OK (some, WIP, are still skipped).

I would go ahead and merge this, suggesting we do not invest more time on the present CI and rather move to using the Azure pipelines as for the `Particle` package. I can open an "issue" and merge this if you are in agreement.

Thanks.

henryiii:
Looks fine. We'll probably set the testing up on Azure so quite a bit of that will probably change.

PR #21, opened 2019-04-02
Add new DecFileParser class, with test suite

eduardo-rodrigues:
With this new `DecFileParser` class I was able to trivially parse and check the master LHCb DECAY.DEC file:

```
from decaylanguage.dec.dec import *
import time

start = time.time()
p = DecFileParser('decaylanguage/data/DECAY_LHCB.DEC')
p.parse()
end = time.time()
print('Parsing time = %s seconds.' % (end-start))
```

The massive file is nicely parsed in under 2 seconds :-).

And, BTW, I found a little bug in the file - duplication of decay definition for the _Sigma(1775)0_ particle! LHCb simulation group about to be informed ...


eduardo-rodrigues:
I will be adding a bit more doc, in particular. But the code is otherwise up for discussion, @henryiii.

eduardo-rodrigues:
@henryiii, this is now ready for review, though I will still add some more documentation, and am also thinking about an extra method to display a la dot-file a decay chain, for example.

henryiii:
Can you rebase to see if this starts passing? Looks good otherwise. I might recommend a `__slots__` just for clarity.

eduardo-rodrigues:
Thanks a lot for the look and the improvements/fixes.

I've just rebased.

eduardo-rodrigues:
Forgot to say: I do want to add more docs, and tests, though I put a significant amount of it already. I will merge, then, as this PR is rather large and contains some nice additions.

Thanks.

PR #25, opened 2019-04-05
New test decay files

eduardo-rodrigues:
Adding a handful of .dec decay files for testing purposes.

@henryiii, what's your take on the location of such files? They are under `decaylanguage/data` but we could as well have then under a new directory `tests/data`.


henryiii:
Testing files should be in tests, since you don't want to distribute them in the package. If something might be of general use (including a simple example file and/or a full file like the LHCb one), that can be in the package in data.

eduardo-rodrigues:
That's my opinion too. All test files moved to new directory `tests/data`.

PR #26, opened 2019-04-06
CHANGELOG update

eduardo-rodrigues:
I put in some more structure in the CHANGELOG for the next release, as it will be a major one.
Let me know if you like it, or if you would prefer something else. Thanks.


eduardo-rodrigues:
Nope. I would release once the decay file parsing code is more complete and robust. Maybe in a week or so.

PR #29, opened 2019-04-12
DecFileParser - more functionality, doc and tests

eduardo-rodrigues:


eduardo-rodrigues:
FYI, @henryiii, there is still one flaw in the logic of the addition of CP conjugate decays. I will get this fixed soon … The rest is ready for review. It should contain all building blocks of the decay file parser.

decaylanguage/dec/dec.py
```diff
@@ -368,16 +808,16 @@ def get_definitions(parsed_file):
         raise RuntimeError("Input not an instance of a Tree!")
 
     try:
-        return {tree.children[0].children[0].value:float(tree.children[1].children[0].value)
-            for tree in parsed_file.find_data('define')}
+        return sorted({tree.children[0].children[0].value:float(tree.children[1].children[0].value)
+            for tree in parsed_file.find_data('define')})
     except:
```
henryiii: Are you sure you want this converted to a list of the keys? If so, don't make a dict. If not, you'll need an OrderedDict and some way to sort by keys.

eduardo-rodrigues: I think it makes sense for the "Define" statements to be available as a dictionary. There is no real need to sort by key but I thought I would do this as nicer for the user to see it all sorted, though I realise this only makes sense in recent Python versions (users of older version won't see the sorting, which anyway is not stated in the doc).
This can easily be changed but I would not complicate it more, for what it is.

Drop an alternative if you feel stronger about this ;-).

henryiii:
```python
>>> sorted({'key': 'value'})
['key']
```

That happens on all versions of Python.

henryiii: If you want this sorted, you have to do so before it becomes a dictionary

eduardo-rodrigues: Just for the record: as decided when we last chatted, I will remove the sort.

PR #30, opened 2019-04-19
Remove notebooks/PDGID.ipynb

eduardo-rodrigues:
Seems @henryiii that you committed this test notebook by mistake ;-).


henryiii:
In general, I need to be very careful about notebooks in these repos. They currently don't get run by default and therefore don't track with updates.

PR #37, opened 2019-05-14
DecFileParser works nicely for master DECAY.DEC

eduardo-rodrigues:
The parsing now produces on the fly the correct/expected number of charge-conjugate decay trees. Also, this is now done much faster - DECAY.DEC is parsed in about 2 seconds.

This should be the last work on the parser before moving on to a design of the "backend storage" for decay chains, as per our discussions. That will be a major improvement and redesign of various parts.

### DIFF
--- a/decaylanguage/dec/dec.py
+++ b/decaylanguage/dec/dec.py
@@ -30,11 +30,14 @@ from __future__ import absolute_import, division, print_function
 
 import os
 import warnings
+import re
 
 from lark import Lark
 from lark import Tree, Transformer, Visitor
 
 from particle import Particle, ParticleNotFound
+from particle.particle.enums import Charge, Charge_undo, Charge_mapping
+from particle.particle.regex import getdec
 
 from ..data import open_text
 from .. import data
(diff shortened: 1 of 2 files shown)


decaylanguage/dec/dec.py
```diff
@@ -369,26 +372,47 @@ def _add_charge_conjugate_decays(self):
...
-        # if mother_names_ccdecays = ['anti-M1', 'anti-M2'],
-        # the relevant Trees are the ones describing the decays of ['M1', 'M2'].
+        # if mother_names_ccdecays = ['anti-M10', 'anti-M2+'],
+        # the relevant Trees are the ones describing the decays of ['M10', 'M2-'].
         dict_cc_names = self.dict_charge_conjugates()
 
-        i = 0
+        name2treepos = {}  # match name -> position in list self._parsed_decays
+        for i, t in enumerate(self._parsed_decays):
+             name2treepos[t.children[0].children[0].value] = i
```
henryiii: Instead of these three lines, you can write a dictionary comprehension:

```python
name2treepos = {t.children[0].children[0].value:i for i, t in enumerate(self._parsed_decays)}
```

eduardo-rodrigues: Good point. Will commit that in a sec ...

PR #39, opened 2019-05-15
Renaming decaylanguage.decay to decaylanguage.modeling

eduardo-rodrigues:
As per our discussion, I'm moving all model- and converters-related stuff from the decay to the modeling submodule.

Once that's done and merged, I will start populating a new decay submodule, which will contain our "engine" for a general representation of decay chains, a core feature.


henryiii:
Looks like tests still have some mentions of "decay".

eduardo-rodrigues:
I should have run the tests before committing. Indeed I forgot 2 little things.

eduardo-rodrigues:
Forgot to say the obvious: let me know if you prefer another name than `modeling` for the submodule. Otherwise we're good.

henryiii:
Let me solicit input, I'll get back to you tomorrow.

henryiii:
@tevans1260, what you do you think?

One other idea could be `amplitudes` or something like that.

eduardo-rodrigues:
As agreed, I will now merge this. We can always renamethe submodule again if thought to be best.

PR #40, opened 2019-05-17
DecayChainViewer class

eduardo-rodrigues:


decaylanguage/decay/__init__.py
```diff
@@ -0,0 +1,3 @@
+from __future__ import absolute_import
+
+from .viewer import DecayChainViewer
```
henryiii: This adds a hard dependency on pydot.

eduardo-rodrigues: Fair enough. Line removed.

PR #42, opened 2019-05-19
DaughtersDict class

eduardo-rodrigues:
This is the first class for the design of a digital representation of full decay chains.
Take it as a building block. It's a convenient way to store final states and make it easy to count all species of particles across decay chains.

### DIFF
--- a/decaylanguage/decay/decay.py
+++ b/decaylanguage/decay/decay.py
@@ -27,7 +27,7 @@ class DaughtersDict(Counter):
         dd = DaughtersDict({'K+': 1, 'K-': 2, 'pi+': 1, 'pi0': 1})
 
         >>> # Constructor from a list of particle names
-        >>> dd = DaughtersDict.from_list(['K+', 'K-', 'K-', 'pi+', 'pi0'])
+        >>> dd = DaughtersDict(['K+', 'K-', 'K-', 'pi+', 'pi0'])
 
         >>> # Constructor from a string representing the final state
         >>> dd = DaughtersDict('K+ K- pi0')
@@ -35,13 +35,6 @@ class DaughtersDict(Counter):
         if iterable and isinstance(iterable, str):
             iterable = iterable.split()
         super(DaughtersDict, self).__init__(iterable, **kwds)
-
-    @classmethod
-    def from_list(cls, daughters):
-        """
-        Constructor from a list of particle names.
-        """
-        return cls({k:daughters.count(k) for k in set(daughters)})
 
     def to_string(self):
         """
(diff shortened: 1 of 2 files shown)


eduardo-rodrigues:
OK, ready for review (likely that I will add a bit of code to this class, but version v1 is certainly here).

decaylanguage/decay/decay.py
```diff
@@ -0,0 +1,77 @@
...
+        This is general *not* the number of dictionary elements.
+        """
+        return sum(n for n in self.values())
+
+    def __add__(self, other):
+        """
+        Add two final states, particle-type-wise.
+        """
+        dd = super(DaughtersDict, self).__add__(other)
+        return DaughtersDict(dd)
```
henryiii:
```suggestion
        return self.__class__(dd)
```

I'm guessing the original class didn't do this, and that's why it returned a Counter instead of a DaughtersDict. Let's not propagate the same mistake. ;)

decaylanguage/decay/decay.py
```diff
@@ -0,0 +1,77 @@
...
+
+        >>> # Constructor from a string representing the final state
+        >>> dd = DaughtersDict('K+ K- pi0')
+        """
+        if iterable and isinstance(iterable, str):
+            iterable = iterable.split()
+        super(DaughtersDict, self).__init__(iterable, **kwds)
+
+    @classmethod
+    def from_list(cls, daughters):
```
henryiii: Is there a method of Counters that does this already?

decaylanguage/decay/decay.py
eduardo-rodrigues: Fair point. Done with next commit.

decaylanguage/decay/decay.py
eduardo-rodrigues: Indeed. Method hence removed. Thanks.

decaylanguage/decay/decay.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        Default constructor.
+
+        Examples
+        --------
+        >>> # An empty final state
+        >>> dd = DaughtersDict()
+
+        >>> # Constructor from a dictionary
+        dd = DaughtersDict({'K+': 1, 'K-': 2, 'pi+': 1, 'pi0': 1})
```
henryiii:
```suggestion
        >>> dd = DaughtersDict({'K+': 1, 'K-': 2, 'pi+': 1, 'pi0': 1})
```

decaylanguage/decay/decay.py
```diff
@@ -0,0 +1,68 @@
...
+    def __str__(self):
+        return repr(self)
+
+    def __len__(self):
+        """
+        Return the length, i.e. the number of final-state particles.
+
+        Note
+        ----
+        This is general *not* the number of dictionary elements.
```
henryiii:
```suggestion
        This is generally *not* the number of dictionary elements.
```

PR #43, opened 2019-05-20
DecayMode class

eduardo-rodrigues:
This is the second class for the design of a digital representation of full decay chains.
Take it as the remaining building block. It's a convenient way to store decay modes, making use of the new DaughtersDict class. The mother particle is not part of / stored in the decay mode as that's common to various decay modes, by construction - no need to duplicate info.

I will need to add some more documentation. Otherwise that's it. In particular, the test suite seems comprehensive already.

### DIFF
--- a/tests/decay/test_decay.py
+++ b/tests/decay/test_decay.py
@@ -97,7 +97,7 @@ def test_DecayMode_describe_with_extra_info():
 def test_DecayMode_string_repr():
     dd = DaughtersDict('p p~ K+ pi-')
     dm = DecayMode(1.e-6, dd, model='PHSP')
-    assert dm.__str__() == "<DecayMode: daughters=K+ p pi- p~, BF=1e-06>"
+    assert str(dm) == "<DecayMode: daughters=K+ p pi- p~, BF=1e-06>"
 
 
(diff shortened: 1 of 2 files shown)


eduardo-rodrigues:
It's now ready for review.

tests/decay/test_decay.py
```diff
@@ -36,3 +37,70 @@ def test_DaughtersDict_add():
...
+    dm = DecayMode(1.e-6, dd, model='PHSP', study='toy', year=2019)
+    assert 'Extra info:' in dm.describe()
+    assert 'study: toy' in dm.describe()
+    assert 'year: 2019' in dm.describe()
+
+
+def test_DecayMode_string_repr():
+    dd = DaughtersDict('p p~ K+ pi-')
+    dm = DecayMode(1.e-6, dd, model='PHSP')
+    assert dm.__str__() == "<DecayMode: daughters=K+ p pi- p~, BF=1e-06>"
```
henryiii:
```suggestion
    assert str(dm) == "<DecayMode: daughters=K+ p pi- p~, BF=1e-06>"
```

PR #47, opened 2019-06-13
Add Charge conjugate methods to decay-related classes

eduardo-rodrigues:


setup.py
```diff
@@ -78,6 +78,7 @@ def get_version():
         'pathlib2>=2.3; python_version<"3.5"',
         'enum34>=1.1; python_version<"3.4"',
         'importlib_resources>=1.0; python_version<"3.7"',
+        'cachetools',
```
henryiii: Let me see if I can do this by using the stdlib version if the user is on Python 3.

eduardo-rodrigues: Else we can also remove the caching for now, as it is not a game-changer, at least for now. I have no strong feeling here.

This being said, I may use cachetools in other places soon ...

henryiii: If you use it in other places, we can move this to a utility file, and then make it a lambda so you can call `cacher(64)` and that will return the correct object to make into a decorator. (Assuming this works)

eduardo-rodrigues: Yes, something to have in mind.

Issue #46, opened 2019-05-23
Deal with variable definitions for decay models in .dec files


eduardo-rodrigues:
Some .dec decay files may define decay model parameters via variables. For the sake of example, the line

`0.000030     J/psi  omega             SVV_HELAMP PKHminus PKphHminus PKHzero PKphHzero PKHplus PKphHplus;`

in `DECAY_LHCB.DEC` uses variables defined such as
`Define PKHminus 0.612`
This is presently not dealt with by the parser, meaning the parser does not check whether model parameters that are not numbers have a matching `Define` statement, in which case it should replace the names by the actual numbers.

eduardo-rodrigues:
Closed via https://github.com/scikit-hep/decaylanguage/pull/48.

PR #48, opened 2019-06-13
Fix parsing of various types of decay model definitions

eduardo-rodrigues:
The fix to the Lark grammar file is subtle in the sense that both "VSS" and ""VSS_BMIX" exist, and the parser needs to look for the latter before the former as otherwise it concludes that for 
```
Decay Upsilon(4S)
1.0    B0    anti-B0       VSS_BMIX dm;
Enddecay
```
the decay model is "VSS", with "_BMIX" a model parameter.

This PR does not yet address https://github.com/scikit-hep/decaylanguage/issues/46, which will be the subject of another PR once this one is in. A test for the above will go in this future PR.


eduardo-rodrigues:
I decided to deal with issue https://github.com/scikit-hep/decaylanguage/issues/46 in the end, as the required code is very much related to the code of the first part of this PR.

eduardo-rodrigues:
Closes https://github.com/scikit-hep/decaylanguage/issues/46.

henryiii:
Haven't looked at it yet, but shouldn't it only be matching complete words?

eduardo-rodrigues:
What do you mean?

eduardo-rodrigues:
Oh, you mean for the .lark file update? I reckon the issue is coming from the default parser, which tries and be fast, hence takes shortcuts where "possible". The update removes possible ambiguities.

henryiii:
I think if we remove the ignore whitespace line, then add explicit whitespace to the parts of the model where whitespace is allowed, then it would match complete words only (as long as we do 1 or more whitespace chars for whitespace). For now, this is fine, and it might be (probably is?) faster.

decaylanguage/dec/dec.py
```diff
@@ -155,6 +155,14 @@ def parse(self, include_ccdecays=True):
         # At last, find all particle decays defined in the .dec decay file ...
         self._find_parsed_decays()
 
+        # Check whether certain decay model parameters are defined via
+        # variable names with actual values provided via 'Define' statements,
+        # and perform the replacements name -> value where relevant.
+        # Do also a replacement of 'a_float' with a_float.
```
henryiii: > Do also a replacement of 'a_float' with a_float.

Is this being done here? It's rather hidden if it is (it's fine if it is).

decaylanguage/dec/dec.py
```diff
@@ -155,6 +155,14 @@ def parse(self, include_ccdecays=True):
...
+        dict_define_defs = self.dict_definitions()
+        [DecayModelParamValueReplacement(define_defs=dict_define_defs).visit(tree)
+        for tree in self._parsed_decays]
```
henryiii: Please make this an explicit loop, rather than relying on the fact that it has to run through the loop to make the list. It's best to not use comprehensions if you don't want the output result. Plus it gives you nice indentation for the two lines for free. (Sometimes it is handy on the REPL to do it this way, since it can avoid an extra line)

decaylanguage/dec/dec.py
```diff
@@ -612,6 +620,59 @@ def __str__(self):
...
+    """
+    def __init__(self, define_defs=dict()):
+        self.define_defs = define_defs
+
+    def _replacement(self, t):
+        try:
+            t.children[0].value = float(t.children[0].value)
+        except AttributeError:
+            try:
+                t.value = self.define_defs[t.value]
```
henryiii: Can you make this an if statement instead of making it a catch all statement?

```python
if t.value in self.define_defs:
    t.value = self.define_defs[t.value]
```

You could also catch just KeyErrors or something like that. `except:` is frowned upon.

Fun fact: Python 3.8+ syntax:
```python
if (val := t.value) in self.define_defs:
    t.value = self.define_defs[val]
```

decaylanguage/dec/dec.py
eduardo-rodrigues: Yep, it is.

decaylanguage/dec/dec.py
eduardo-rodrigues: Sure.

decaylanguage/dec/dec.py
eduardo-rodrigues: Fair enough.

Issue #50, opened 2019-06-19
Add handy class constructors for DecFileParser


eduardo-rodrigues:
This is really handy if one wants to instantiate the parser with a user decay *together* with all generic decays defined in a master DECAY.DEC. In fact this is likely to be the most useful instantiation.

eduardo-rodrigues:
Two enhancements come immediately to my mind:
```
DecFileParser.from_string(<file given as string>)
DecFileParser.from_files('DECAY.DEC', 'my_decay.dec')
```

eduardo-rodrigues:
Closed via https://github.com/scikit-hep/decaylanguage/pull/51.

PR #51, opened 2019-06-19
New class constructors for DecFileParser

eduardo-rodrigues:
Addresses https://github.com/scikit-hep/decaylanguage/issues/50.

It also addresses trouble found by @dbobulsk parsing her user files.

### DIFF
--- a/decaylanguage/dec/dec.py
+++ b/decaylanguage/dec/dec.py
@@ -32,6 +32,8 @@ import os
 import warnings
 import re
 
+from six import StringIO
+
 from lark import Lark
 from lark import Tree, Transformer, Visitor
 
@@ -65,17 +67,18 @@ class DecFileParser(object):
 
     Example
     -------
-    >>> parsed_file = DecFileParser.from_file('my-decay-file.dec')
+    >>> parsed_file = DecFileParser('my-decay-file.dec')
     """
 
     __slots__ = ("_grammar",
                  "_grammar_info",
-                 "_dec_file_name",
+                 "_dec_file_names",
+                 "_dec_file",
                  "_parsed_dec_file",
                  "_parsed_decays",
                  "_include_ccdecays")
 
-    def __init__(self, filename):
+    def __init__(self, filename=None):
         """
         Default constructor.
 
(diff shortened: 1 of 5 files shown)


eduardo-rodrigues:
@henryiii, this is ready from my side.

decaylanguage/dec/dec.py
```diff
@@ -87,33 +90,78 @@ def __init__(self, filename):
...
+        _cls._dec_file = stream.read()
 
-        # Check input file
-        if not os.path.exists(filename):
-            raise FileNotFoundError("'{0}'!".format(filename))
+        return _cls
 
-        return cls(filename)
+    @classmethod
+    def from_files(cls, filenames):
```
henryiii: Couldn't either a) call the base constructor here with each file and merge, or b) allow a list of decay files in the base constructor?

decaylanguage/dec/dec.py
```diff
@@ -65,17 +67,18 @@ class DecFileParser(object):
...
                  "_grammar_info",
-                 "_dec_file_name",
+                 "_dec_file_names",
+                 "_dec_file",
                  "_parsed_dec_file",
                  "_parsed_decays",
                  "_include_ccdecays")
 
-    def __init__(self, filename):
+    def __init__(self, filename=None):
```
henryiii: This could be `self, *filenames`

henryiii: The main constructor body would pretty much look just like the current from_files body.

eduardo-rodrigues:
Yes, could be a good thing. I just need to see how to work that out together with the `from_string` class constructor, which also seems handy, hence worth keeping.

decaylanguage/dec/dec.py
eduardo-rodrigues: I've removed this class constructor making the default one versatile.

decaylanguage/dec/dec.py
eduardo-rodrigues: Yep, done.

dbobulsk:
Many thanks for these new features which also address my issue, I'm gonna give it a try :)

PR #60, opened 2019-07-18
DecayChainViewer made more robust

eduardo-rodrigues:


eduardo-rodrigues:
Merging since the failure for Linux + Python 2.7 is a different painful issue, totally unrelated.

Issue #66, opened 2019-08-28
JetSetPar not recognized?


ch2ohch2oh:
There are lines starting with JetSetPar in my decfile which caused the package to throw a `UnexpectedToken` exception. Such lines are not relevant to the visualization but should be handled.

eduardo-rodrigues:
Thanks for using the package, @ch2ohch2oh!

Indeed all dec files are from LHCb as those are the ones I have at hand. It can well  happen that they do not contain all features available/usable by EvtGen, hence the unkown token, i.e. keyword. That's an easy fix, which I will do asap. In that respect, can you provide me with your dec file? There is nothing private about those, so no worries.

Checking the code I see that all JetSetPar statements are commented out, hence unknown to the package so far. Also the model SSS_CP needs to be added.

BTW, the default master dec file we load is the LHCb one. I would be more than happy to add to the repository the latest and greatest Belle2 DECAY.DEC if anyone could send it to me … It would be rather useful, actually.

Thanks!

eduardo-rodrigues:
For reference, let me explain how to trivially deal with the issue of the unknown model name (it should not prevent us from updating the package to know about it!). It is just a matter of adding the model name to the list in `decaylanguage/data/decfile.lark`, see the line

`MODEL_NAME.2 : "BaryonPCR"|"BTO3PI_CP"|"BTOSLLALI"|...`

Then proceed as usual apart from adding an extra line to call to `load_grammar` to specific the Lark grammar to use :-):

```
p = DecFileParser('my_decfile.dec')
p.load_grammar('my_updated_decfile.lark')
p.parse()
...
```

eduardo-rodrigues:
Brilliant, it will be very useful.

What about your user decfile with which you had trouble? I'm pretty sure the PR I created earlier today solves the problems you mention, but better be maniac ;-).

eduardo-rodrigues:
You mean the master decfile? That's not crazy. It's fair.

OK, I will do some tests as I won't be surprised if Belle 2 decfile contains some extra models we don't, and I want to put them in.

eduardo-rodrigues:
I will probably finish tomorrow as it's getting late. In any case there is a buglet at line 4142 in that it is

`0.0030      D_s+        pi-                   PHOTOS PHOTOS PHSP; `

[...]

PR #65, opened 2019-08-26
Decay chain DOT graphs display HTML particle names

eduardo-rodrigues:

PR #67, opened 2019-08-28
DecFileParser extension to understand JETSET definitions

eduardo-rodrigues:
Fixes https://github.com/scikit-hep/decaylanguage/issues/66.


eduardo-rodrigues:
The annoying error for Python 2.7 has already been dealt with / fixed in PR https://github.com/scikit-hep/decaylanguage/pull/65 that is about to be merged. So let's ignore the issue here. It seems, BTW that it's got to do with Lark updates as it was not present before.

PR #69, opened 2019-08-29
Improvement on how decfile parsed JETSET parameters are stored

eduardo-rodrigues:


decaylanguage/dec/dec.py
```diff
@@ -1140,16 +1140,30 @@ def get_jetset_definitions(parsed_file):
...
+        Trivial helper function to convert the parsed (as strings)
+        JETSET parameters into what they are, namely integers or floats.
+        """
+        try:
+            return int(n)
+        except ValueError:
+            try:
+                return float(n)
+            except:
+                # handle all other exceptions
```
henryiii:
```suggestion
                # pass though non-numbers unchanged
```

PR #72, opened 2019-10-08
DecayChain class

eduardo-rodrigues:
This is the third and main class for the design of a digital representation of full decay chains.
It's a convenient way to store an entire decay chain/tree, making use of the DecayMode class.


eduardo-rodrigues:
Hi @henryiii, I finally got back to this work. I still have much planned but this class is a central piece … One can also visualise a `DecayChain` using the `to_dict()` method and pass that to `DecayChainViewer` :-).

PR #79, opened 2019-10-18
Add new models from Belle2

eduardo-rodrigues:


eduardo-rodrigues:
FYI @daritter.

daritter:
just for my education: Is the list of decay models in the lark file auto generated from the enums.py? Or do you have a unit test to make sure those two lists match?

eduardo-rodrigues:
None of the two. But that is something to add, which I had forgotten in the meantime. Will do asap.
Reason for not auto-generating is that the lark file should be standalone as the user can provide its own. I will add a consistency check. Thanks for the reminder :-)!

Issue #87, opened 2019-11-08
Binder Fails


maluethi:
Just discovered that when I launch binder (trough the icon on the main github page), I get an error during the machine build.
```
Pip subprocess error:
ERROR: Invalid requirement: 'particle=0.6.*' (from line 2 of /home/jovyan/condaenv.lrqydel1.requirements.txt

Hint: = is not a valid operator. Did you mean == ?
CondaEnvException: Pip failed
```

eduardo-rodrigues:
Thanks for reporting. I will check this asap as I'm presently on holidays … The hint seems correct, yes.
(Strange as I was going to swear that I checked the button recently, when updating the YAML. Seems not.)

eduardo-rodrigues:
@maluethi, this is now fixed with PR https://github.com/scikit-hep/decaylanguage/pull/88. Thanks again for reporting.

Note that you can also have a look at my repo https://github.com/eduardo-rodrigues/2019-10-16_PyHEP2019Workshop for a recent and rather comprehensive overview of `DecayLanguage` and `Particle` (I should update the demos using material in this presentation).

Just out of interest: what's your use-case for looking into `DecayLanguage`? Did you bump into this package by chance or ...? It's really difficult to know the users of our packages, especially if we get no stars. But it is super useful to get some insight :-).

maluethi:
Hi Eduardo. At the moment I am not actively using your package. I found while browsing through scikit-hep.org and looking at packages that would be useful for my analysis. A while ago I implemented something similar (very immature) for some analysis. It is nice to know someone did the hard work already and I might use the package when the analysis is a bit more grown up.

eduardo-rodrigues:
OK, sounds good, @maluethi. Thanks.

PR #88, opened 2019-11-14
Trivial env fix for Binder

eduardo-rodrigues:


eduardo-rodrigues:
Fixes https://github.com/scikit-hep/decaylanguage/issues/87.

Issue #90, opened 2019-12-27
An extra '2' in one line of LHCB dec file?


ch2ohch2oh:
https://github.com/scikit-hep/decaylanguage/blob/9a02b51fe2db7abf001c8a81d690a1f9e6d14857/decaylanguage/data/DECAY_LHCB.DEC#L291

There seems to be an extra digit 2 in one line of the dec file. It should work with current version of the `decfile.lark` but I believe it should not be there. I was playing around with the `decfile.lark` to incorporate decay line like this
```
Decay xxx
0.5 SOME_MODEL 1 2 3
      4 5 6; # some comments
Dnddecay
```
That is how I found this extra 2.

eduardo-rodrigues:
First of all, many thanks for using the package!

Yes, this seems to be a buglet in the LHCb master file and I will pass the information on. I will check what information gets returned after parsing, since the parsing does end OK.

ch2ohch2oh:
From my understanding, the program parsed this line fine because in your `decfile.lark`, both `;` and `#` are treated as keywords for comments.

eduardo-rodrigues:
Yep, indeed. I will try and review your related PR today, to close this.

eduardo-rodrigues:
Closed via https://github.com/scikit-hep/decaylanguage/pull/91.

henryiii:
Have you verified that this has been changed in the official file? This file is only an example, the goal is to read the real file, not a modified copy.

eduardo-rodrigues:
Did you not see my last comment in https://github.com/scikit-hep/decaylanguage/pull/91? Does it not directly address your worry? Unless I misunderstood. The removal of a comment letter is not a reason to block the PR, I reckon.

henryiii:
Okay, as long as the problem gets fixed, I'm fine with it being strict. It would be nice if we can get this used officially, yes. I'd like users of Decfiles to also be stricter, though.

eduardo-rodrigues:
Well, that's the idea, of course.

As for stricter rules, which make DecFiles better, that would be great. But keep in mind that the "showstopper" is EvtGen, which is the consumer of DecFiles. Clearly EvtGen is sloppy in various places. A comprehensive fix would have to be propagated there too. Fortunately the LHCb coordinator I talked to is one of the EvtGen maintainers ;-).

PR #91, opened 2019-12-27
Belle II dec file and small fixes

ch2ohch2oh:
1. Removed the extra `2` in one line of LHCB dec file
2. Added BELLE II dec file and a simple test
3. Changed `Particles.from_dec` to `Particles.from_string` since this breaks the tests
4. Small change to `decfile.lark` to support "multiline" decay
```
Decay xxxx
0.123 SOME_MODEL 1 2 3
                             4 5 6; # this model has 6 parameters
Enddecay
```


decaylanguage/data/decfile.lark
```diff
@@ -3,7 +3,7 @@
 // Distributed under the 3-clause BSD license, see accompanying file LICENSE
 // or https://github.com/scikit-hep/decaylanguage for details.
 
-start : _NEWLINE? (line _NEWLINE)+ ("End" _NEWLINE)?
+start : _NEWLINE* (line _NEWLINE+)+ ("End" _NEWLINE+)?
```
eduardo-rodrigues: Why a `+` in the `_NEWLINE+` after `End`?

decaylanguage/dec/dec.py
```diff
@@ -452,7 +452,7 @@ def _add_charge_conjugate_decays(self):
         def _is_not_self_conj(t):
             try:
                 mname = t.children[0].children[0].value
-                if Particle.from_dec(mname).is_self_conjugate:
+                if Particle.from_string(mname).is_self_conjugate:
```
eduardo-rodrigues: In the end all `from_dec` are in fact to be replaced by `from_evtgen_name` but there are some cases missing in the `Particle` package, which explains the failures you probably saw in your tests, which made you use `from_string`. I've had since weeks a start of code to fix all this. It's now high on my list so that this package can finally depend on `Particle` 0.8.*. For now let's go with your temporary workaround (indeed `from_string` is not guaranteed to always return the correct match).

tests/dec/test_dec.py
```diff
@@ -411,6 +411,13 @@ def test_master_DECAYdotDEC_file():
 
     assert p.number_of_decays == 506
 
+def test_BELLE2_decfile():
+    p = DecFileParser(DIR / '../../decaylanguage/data/DECAY_BELLE2.DEC')
+    p.parse()
+
+    # Just check the dec file will parse since I do not know 
+    # how many decays are in the dec file.
+    assert p.number_of_decays > 0
```
eduardo-rodrigues: You do know precisely how many decays are in the file, once you parse successfully. I reckon it is fine to just check for that number in the test, knowing that the test is likely to need an update if the dec file gets updated.

decaylanguage/data/decfile.lark
henryiii: That would allow End followed by multiple newlines, but what is after End doesn't matter anyway IIRC, so I think it could safely go either way.

decaylanguage/dec/dec.py
```diff
@@ -848,9 +848,9 @@ def find_charge_conjugate_match(pname, dict_cc_names=dict()):
 
     # Search EvtGen names via the Particle class
     try:
-        return Particle.from_dec(pname).invert().name
+        return Particle.from_string(pname).invert().name
     # If anything else fails ...
-    except ParticleNotFound:
```
henryiii: It's generally a bad idea to use `except:`. Can't you list the possible exceptions?

henryiii: This also [catches Control C and system exit!](http://effbot.org/zone/stupid-exceptions-keyboardinterrupt.htm) If you *really* have to catch all, you should at least exclude `KeyboardInterrupt, SystemExit`, but it's *much* better to list the possible exceptions instead.

decaylanguage/data/decfile.lark
eduardo-rodrigues: Indeed, since empty lines are ignored. If anything, without the + the parsing will be an epsilon faster.

tests/dec/test_dec.py
henryiii: Can't you run this once, get the number, then set it here? Then if we fix/improve/update the file, it will change and need to be changed here too, but at least we will know when it changes.

henryiii:
I see you need it for multiline decays. Okay, I think that's a good reason to make it work correctly with the semicolons; we could add a strip-after-simicolon feature or preprocessing to avoid changing the LHCb dec file for now. (I'd really love it if that was fixed officially, though).

tests/dec/test_dec.py
eduardo-rodrigues: I made the same comment above. I do check the correct number in a similar test of the LHCb master file.

eduardo-rodrigues:
@henryiii, I had a chat with one of our LHCb simulation coordinators just before Christmas and we are meant to discuss with more time now in January. The discussion was about a possible use of this package in LHCb. This buglet/feature is yet another reason to get the package used in LHCb and I will make sure the issue found here gets known and hopefully corrected, see my comment at https://github.com/scikit-hep/decaylanguage/issues/90.

decaylanguage/dec/dec.py
eduardo-rodrigues: Agreed. Better extend the exception instance I had listed.

tests/dec/test_dec.py
ch2ohch2oh: Oh I see, it is a guard for future changes. Thanks for pointing this out.

eduardo-rodrigues:
Hi @ch2ohch2oh, can you please get in touch with me privately (I'm sure you know my email) to have a proper reference for example for the new major release 0.6.0 just prepared (https://zenodo.org/record/3603847#.XhhL8W5FxPY)? Thanks.

PR #94, opened 2020-01-09
DecayMode and DaughtersDict classes - enhancements and more doc

eduardo-rodrigues:


eduardo-rodrigues:
@henryiii, this is probably my last PR for the next major release, which is due since a while as several major releases of `Particle` came out without a related release of `DecayLanguage`. I will merge this later unless you worry about any of it ... For sure it's a step further in a few update I have in mind to make the handling of decay files more robust and versatile.

Issue #99, opened 2020-06-02
pydot required but not installed


daritter:
I just installed decaylanguage via `pip3 install decaylanguage` and then tried to

```
from decaylanguage import DecFileParser
```
but this resulted in

```
Traceback (most recent call last):
  File "/home/ritter/belle2/software/decfiles/tests/parse_decfiles.py", line 2, in <module>
    from decaylanguage import DecFileParser
  File "/home/ritter/belle2/externals/v01-09-00/Linux_x86_64/common/lib/python3.6/site-packages/decaylanguage/__init__.py", line 14, in <module>
    from .decay import DecayChainViewer
  File "/home/ritter/belle2/externals/v01-09-00/Linux_x86_64/common/lib/python3.6/site-packages/decaylanguage/decay/__init__.py", line 4, in <module>
    from .viewer import DecayChainViewer
  File "/home/ritter/belle2/externals/v01-09-00/Linux_x86_64/common/lib/python3.6/site-packages/decaylanguage/decay/viewer.py", line 21, in <module>
    raise ImportError("You need pydot for this submodule. Please install pydot with for example 'pip install pydot'\n")
ImportError: You need pydot for this submodule. Please install pydot with for example 'pip install pydot'
```

It seems that pydot is not optional but apparently not in the requirements. Can we make it an optional dependency? Or add it to the requirements?

henryiii:
Oops, yes, one of the two. Probably an optional requirement; we need to fix the error when not using the viewer, and probably add a better `[vis]` or `[all]` or something like that optional requirement in the pip install.

eduardo-rodrigues:
The message is clear - "You need pydot for this submodule. Please install pydot with for example 'pip install pydot'" ;-). But agreed that we can expand the set up. I have no strong opinion. Right now you get `pydot` with [notebook], for example, see `https://github.com/scikit-hep/decaylanguage/blob/master/setup.py`.

eduardo-rodrigues:
BTW, before you ask - I know that we need to make a new release to patch the major release 0.10.0 of `Particle` … It's on the to-do list.

henryiii:
`from decaylanguage import DecFileParser` should not require pydot, though.

eduardo-rodrigues:
Yeah, that's because the class is available from the top level, see `https://github.com/scikit-hep/decaylanguage/blob/master/decaylanguage/__init__.py`.

daritter:
[...]

PR #101, opened 2020-06-02
Only export DecayChainViewer if available

henryiii:
DecayChainViewer will not be available at the top level if pydot is not available. See #99.


eduardo-rodrigues:
Seems fair to me Feel free to merge.

PR #102, opened 2020-06-05
Make pydot required, bump version to 0.6.2

eduardo-rodrigues:
Fixes https://github.com/scikit-hep/decaylanguage/issues/99.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Version 0.6.2 (2020-06-05)
+
+* Dependencies:
+  - Package dependency on ``pydot`` made a requirement.
+
+
 ## Version 0.6.1 (2020-01-15)
 
 * Parsing of decay files (aka .dec files):
--- a/decaylanguage/_version.py
+++ b/decaylanguage/_version.py
@@ -4,7 +4,7 @@
 # or https://github.com/scikit-hep/decaylanguage for details.
 
 
-__version__ = '0.6.1'
+__version__ = '0.6.2'
 
 version = __version__
 version_info = __version__.split('.')
--- a/decaylanguage/decay/__init__.py
+++ b/decaylanguage/decay/__init__.py
@@ -1,7 +1,4 @@
 
 from .decay import DaughtersDict, DecayMode, DecayChain
 
-try:
-    from .viewer import DecayChainViewer
-except ImportError:
-    pass
+from .viewer import DecayChainViewer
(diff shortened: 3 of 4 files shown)


henryiii:
I'm confused with the discussion. Why are we making pydot required? We then need to test to see what happens if pydot is available, but the compiled graphviz library is not available, so we can nicely error out.

eduardo-rodrigues:
The alternative is to stay as is, with your patch on the important statement. But then people such as @daritter will still have to do some extra installs - pydot and the compiled grapgviz. And his installation will probably be pip install decaylanguage[notebook]. What's best? I'm just thinking for the user and installations as for example Belle II … I have no strong feeling, to be honest.

henryiii:
Compiled graphviz will always be required. But I am thinking it might be better to do it this way, we should just test to see what happens if we don't have the library installed, to make sure the error message is clear enough.

If we do this, we can remove my check for the library, since it will always be there.

How hard would it be to do the switch to graphviz (Python) since we are messing with it? Do you know how similar the two are?

eduardo-rodrigues:
> Compiled graphviz will always be required. But I am thinking it might be better to do it this way, we should just test to see what happens if we don't have the library installed, to make sure the error message is clear enough.
> If we do this, we can remove my check for the library, since it will always be there.

Now you got me confused and I'm not sure how you would prefer to proceed. So I let you either proceed or close this PR :-).

> How hard would it be to do the switch to graphviz (Python) since we are messing with it? Do you know how similar the two are?

I think I made the comment in another thread that we should eventually remove `pydot` since it's not being maintained. I would rather do it for the next major version, though. To me the important things now are to (1) get a new minor version avoiding the issue Martin raised and then (2) make a new release supporting `Particle` 0.10, to gain again full compatibility (`DecayLanguage` 0.6 rely on `Particle` 0.9). I believe this makes sense.

henryiii:
I don't know if there's that big of an issue missing the compiled graphviz, and pydot/graphviz for python should handle it elegantly/clearly. So I like this solution. I've reverted my check for pydot.

henryiii:
If we do release, can I apply some of the developer changes on the scikit-hep page? Specifically, I'm looking for a package to test the new release mechanism on.

eduardo-rodrigues:
> If we do release, can I apply some of the developer changes on the scikit-hep page? Specifically, I'm looking for a package to test the new release mechanism on.

If that can help you, sure!

eduardo-rodrigues:
@henryiii, shall we then merge this one? Thanks.

henryiii:
Sure, just not ready to release until #103 goes in.

PR #103, opened 2020-06-05
ci: Adding github build and deploy

henryiii:
Still needs:
- [x] Badge on README
- [x] Token in GitHub secrets


henryiii:
Hopefully we can standardize and simplify the package for next time, but for now, this should get close enough that I can test deployment. I'll make a few beta releases to make sure it works.

henryiii:
@eduardo-rodrigues, we are ready to release the next patch version now, correct? I can make a final release once I see that the beta release(s) work.

henryiii:
Actually, we don't have `setuptools_scm` here yet, so I'll just make a "final" 0.6.2 release and then keep updating in GitHub it if it doesn't work. Once it deploys to PyPI it should be fine.

henryiii:
Since you put a date in, I'm assuming this is the only change. Making release.

PR #104, opened 2020-06-05
Dependencies update

eduardo-rodrigues:
Update for dependency on `Particle` was a must. I "refreshed" some other dependencies, while at it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Version 0.7.0
+
+Under development.
+
+* Dependencies:
+  - Package dependent on ``Particle`` versions 0.10.*.
+  - Dependencies on `lark-parser` and others upgraded.
+
+
 ## Version 0.6.2 (2020-06-05)
 
 * Dependencies:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 Under development.
 
 * Dependencies:
-  - Package dependent on ``Particle`` versions 0.10.*.
+  - Package dependent on ``Particle`` version 0.10.
   - Dependencies on `lark-parser` and others upgraded.
 
 
--- a/setup.py
+++ b/setup.py
@@ -19,17 +19,17 @@ from setuptools import setup
 PYTHON_REQUIRES = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 
 INSTALL_REQUIRES = [
-    'attrs>=17.4',
-    'plumbum>=1.6.6',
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## Version 0.7.0
+
+Under development.
+
+* Dependencies:
+  - Package dependent on ``Particle`` versions 0.10.*.
```
henryiii, line 8:
```suggestion
  - Package dependent on ``Particle`` version 0.10.
```
This is making the markdown unhappy, so maybe just make this 0.10, patch versions are implied.

henryiii:
You generally don't want your install requires to be any tighter than it has to be; I would only bump a minimum there if you really have to (Particle). This means people can't install decaylanugage when they have a perfectly good but slightly older versions of a dependency without forcing it to upgrade, too.

setup.py
```diff
@@ -19,17 +19,17 @@
 PYTHON_REQUIRES = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 
 INSTALL_REQUIRES = [
-    'attrs>=17.4',
-    'plumbum>=1.6.6',
+    'attrs>=19.2',
```
henryiii, line 22: I'd restore this, unless we use features in 19.2.

CHANGELOG.md
eduardo-rodrigues: Sure, no prob.

setup.py
eduardo-rodrigues: I should have been explicit here - checking the dependencies of `Particle` I noticed the upgrade there, which is this one. Hence, in practice, we will have to use 19.2 coming from the requirement in `Particle`. I would not go and change the dependency in `Particle` at this stage.

eduardo-rodrigues:
> You generally don't want your install requires to be any tighter than it has to be; I would only bump a minimum there if you really have to (Particle). This means people can't install decaylanugage when they have a perfectly good but slightly older versions of a dependency without forcing it to upgrade, too.
> (We can probably force Lark upgrades, because we are highly coupled to it, and I'm fine bumping plumbum up; so mostly we want to keep attr at the minimum we need.)

See my replies above. In general I totally agree with your comment on avoiding to force upgrades that aren't really required.

setup.py
eduardo-rodrigues: OK, @henryiii ?

henryiii: That’s okay, then; though does DL use attrs? We can just inherit Particle’s if we don’t use it directly

eduardo-rodrigues: Yep, you use it in your classes related to AmpGen.

PR #106, opened 2020-08-13
Bump version to 0.7.0

eduardo-rodrigues:
FYI @henryiii - so it seems that we never released a version compatible with `Particle` v0.10. At this stage, and given that v0.11.0 is best, it makes sense to simply release on the later version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,9 @@
 # Changelog
 
-## Version 0.7.0
-
-Under development.
+## Version 0.7.0 (2020-08-13)
 
 * Dependencies:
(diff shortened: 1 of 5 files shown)


henryiii:
> 'dict' object is not reversible

We should be able to do `reverse(list(dict))`

eduardo-rodrigues:
I suspect there's actually a bug in Lark itself. Am just checking with versions 0.8 rather than 0.9 ...

eduardo-rodrigues:
Indeed there is something fishy but am going forward since we do not need to depend on the very latest Lark. The important thing right now is to be able to release a version in pair with ``Particle 0.11.0.

PR #108, opened 2020-10-05
Don't install "tests" directory

chrisburr:
I noticed in conda-forge/staged-recipes#12786 that the tests/ directory is being installed to site-packages as it isn't explicitly excluded.


eduardo-rodrigues:
Thanks @chrisburr for spotting this.

This reminds me that at some point we should restructure `DecayLanguage` as we did for `Particle` ...

PR #112, opened 2020-11-02
Add tests of Python 3.8 and 3.9 on Windows

eduardo-rodrigues:
At the moment we have a mix of tests. Something to improve at some stage. In any case it is important to have tests for Python 3.8 and 3.9 on Windows.


eduardo-rodrigues:
This is ready, @henryiii.

henryiii (approved):

Issue #100, opened 2020-06-02
ModelAlias support


daritter:
Hi,

it seems our latest decay file is no longer compatible with decaylanguage as we now have ModelAlias lines in our main decay file to define some aliases for models. For example

```
ModelAlias SLBKPOLE_Dtoetaplnu SLBKPOLE 1.0 0.281 1.0 2.010;
```
Supporting the ModelAlias itself is obviously easy enough but then also supporting it to be used as models in the decay files seems currently a bit non-trivial.

eduardo-rodrigues:
Hi @daritter, it's actually rather easy to deal with a so-far unknown decay model on the fly, see the section "advanced usage" I had at some point added to the README. This being said, it's just nicer to continue expanding the known models as they get created. But you folks at Belle have enhanced the master decay file with something new? OK, then we will need some info for the parser …

BTW, it would then be great to update your master decay file at `https://github.com/scikit-hep/decaylanguage/tree/master/decaylanguage/data` for the benefit of everyone.

daritter:
I've seen the README but that was still talking about a fixed list of models, now it seems people on my side started to use a dynamic list of models. And I'm quite sure it's actually not a new feature as we're stuck on a rather old evtgen for the time being so this is something which evtgen has supported for quite some time now :D

As a matter of fact looking in the `EvtGenBase/EvtDecayTable.cpp` there seem to be a few other things supported that decaylanguage doesn't do:

* Define
* ModelAlias
* PythiaAliasParam
* PythiaGenericParam
* Particle
* ChangeMassMin
* ChangeMassMax
* IncludeBirthFactor
* IncludeDecayFactor
* CopyDecay
* RemoveDecay
...

At least some of these are obviously a bad idea to have (Particle (re-)definition in the decay file seems like a recipe for unpleasantness to me) and I'm not demanding that you support all of that. But the ModelAlias actually sounds like a good idea. I've also seen the Change*Mass in one signal decay file of ours but there I'm more inclined on pushing this out of the decay file.

Don't get me wrong, I'm really happy with the effort of decaylanguage. I am mostly making issues so that this is documented. Problem with giving you our decayfile always the same: we haven't managed to get our software licensed yet so I have to ... dance around that

eduardo-rodrigues:
[...]

PR #113, opened 2020-11-04
Make the decay file parser understand the CopyDecay statement

eduardo-rodrigues:
Address some of the possible statements of dec files that the present parser is not aware of, hence it cannot parse, see the discussions at https://github.com/scikit-hep/decaylanguage/issues/100.

FYI @daritter - I found on the web some example master decay files `DECAY.DEC `with which I could work to extend `DecayLanguage`'s parser.

### DIFF
--- a/decaylanguage/data/decfile.lark
+++ b/decaylanguage/data/decfile.lark
@@ -4,7 +4,7 @@
 // or https://github.com/scikit-hep/decaylanguage for details.
 
 start : _NEWLINE* (line _NEWLINE+)+ ("End" _NEWLINE+)?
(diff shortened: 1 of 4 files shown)


eduardo-rodrigues:
PR is WIP because:
1) Make the parser understand `CopyDecay`. DONE.
2) Provide code to return parsed info from `DecFileParser` class and via a standalone function. DONE.
3) Actually make the copy of the decay, under a new name, and store appropriately in `DecFileParser`. TO BE DONE.

eduardo-rodrigues:
All done.

Issue #105, opened 2020-06-09
Replace pydot with graphviz


eduardo-rodrigues:
The package [pydot](https://github.com/pydot/pydot) is no longer maintained, as obvious from the time (Dec. 2018) of the last commit and the increasing number of open issues. We should switch to [graphviz](https://github.com/xflr6/graphviz) in the near future.

Bhanu-mbvg:
I would like to work on this

eduardo-rodrigues:
Great. Looking forward to it.

For you to get started: when forking the repository you will see that `pydot` is used in the `decay/viewer.py`. The idea is to replace all occurrences and related doc with the `graphviz` package, which is maintained. You can use for example the demo notebook with `pydot` or `graphviz` to compare your work and make sure all is OK. Let me know how you get along. Free free to post questions. Thanks.

Bhanu-mbvg:
Hey sure I'll get started thank you for the head's up. Will reach out to you if I face any issues

Bhanu-mbvg:
Hey @eduardo-rodrigues , I'm having little doubts with 'pydot.Dot' and setting default node and edge migrations hope you can help.  

![image](https://user-images.githubusercontent.com/44861163/102641406-09267000-4182-11eb-9747-3a1f21460e70.png)

eduardo-rodrigues:
(I will be slow from now on since starting holidays.)

This bit of code is simply to make sure that a nice set of defaults are used, well, by default, but that the user can still have the flexibility to change some of the settings such as colours. The one thing the user is not allowed to do is change the graph type, which makes sense. What is the trouble? I assume that graphviz is able to deal with defaults for the underlying DOT language graph.

eduardo-rodrigues:
You should look at the defaults I set via the `_get_XX_defaults` methods ...

Does that help or you're having trouble with something else?

eduardo-rodrigues:
BTW, feel free to make a first stab at the work specifying it as draft/WIP and it's then easy to review and iterate.

PR #127, opened 2021-03-24
Replace pydot with graphviz

eduardo-rodrigues:
Indeed `pydot` isn't really maintained since a while. Better be on the safe side.

Closes https://github.com/scikit-hep/decaylanguage/issues/105.


eduardo-rodrigues:
@henryiii, I don't understand how Sourcery is working. It is failing my PR and making a close in 128. Why is that? I would have expected to get suggested changes here, full stop. Hmm.

In any case the PR is ready for review :).

setup.py
```diff
@@ -70,7 +69,7 @@ def get_version():
     install_requires=INSTALL_REQUIRES,
     tests_require=extras["test"],
     extras_require=extras,
-    keywords=["HEP", "particle", "decay", "representation"],
+    keywords=["HEP", "particle", "decay", "chain" "representation"],
```
henryiii, line 72:
```suggestion
    keywords=["HEP", "particle", "decay", "chain", "representation"],
```

This is currently:
```python
    keywords=["HEP", "particle", "decay", "chainrepresentation"],
```

Which I don't think is what you meant.

henryiii:
In the near future I'll adapt to follow the scikit-hep dev guidelines here and in Particle.

setup.py
henryiii: I think our flake8 (at least bugbear) testing would catch this.

eduardo-rodrigues: My eyes did not see the missing comma. Yes, that would be found, I guess too.

eduardo-rodrigues:
> In the near future I'll adapt to follow the scikit-hep dev guidelines here and in Particle.

Indeed this package hasn't yet been updated for all guidelines, and it would be great :-).

eduardo-rodrigues:
Any other comment on this PR, @henryiii? Else I merge. We can try and have the package reorganised (e.g. code under src/) according to guidelines for the next minor release, then, as you suggested above.

henryiii:
Sounds good, I can try to help with some of the modernization after this PR is merged.

henryiii (approved):

eduardo-rodrigues:
> Sounds good, I can try to help with some of the modernization after this PR is merged.

BTW, on that - have you done the modernisation of various packages "by hand" or ...? I can start off if you like.

henryiii:
Setup.py -> setup.cfg I have mostly automated. https://github.com/henryiii/setuptools-modernize

Other than that one, mostly manual.

eduardo-rodrigues:
OK, thanks!

PR #129, opened 2021-03-25
chore: update to setup.cfg

henryiii:
Converting to setup.cfg. Automatic + minor touchup manually.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,5 @@
 requires = [
     "setuptools>=42",
     "wheel",
-    "setuptools_scm[toml]>=3.4"
 ]
 build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,7 @@
+[build-system]
+requires = [
+    "setuptools>=42",
+    "wheel",
+    "setuptools_scm[toml]>=3.4"
+]
+build-backend = "setuptools.build_meta"
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,8 +1,66 @@
+[metadata]
+name = DecayLanguage
+url = https://github.com/scikit-hep/decaylanguage
+author = Henry Fredrick Schreiner III, Eduardo Rodrigues
+author_email = henry.schreiner@cern.ch, eduardo.rodrigues@cern.ch
+maintainer = The Scikit-HEP admins
+maintainer_email = scikit-hep-admins@googlegroups.com
+classifiers =
+    Development Status :: 4 - Beta
+    Intended Audience :: Developers
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: BSD License
+    Operating System :: OS Independent
+    Programming Language :: Python
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Topic :: Scientific/Engineering
+license = BSD 3-Clause License
+description = A language to describe particle decays, and tools to work with them.
+long_description = file: README.md
+long_description_content_type = text/markdown
+keywords =
+    HEP
+    particle
+    decay
+    chain
+    representation
+platforms = Any
+
+[options]
+packages = find:
+install_requires =
+    attrs>=19.2
+    plumbum>=1.6.9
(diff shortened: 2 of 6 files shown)


setup.cfg
```diff
@@ -1,14 +1,72 @@
...
+    decay
+    chain
+    representation
+
+[options]
+packages = find:
+install_requires =
+    attrs>=19.2
+    graphviz
+    lark-parser>=0.8.0,<0.8.6
```
henryiii, line 42: We need to fix this to work with newer versions...

setup.cfg
```diff
@@ -1,14 +1,72 @@
...
+    pandas>=0.22
+    particle==0.14.*
+    plumbum>=1.6.9
+    six>=1.11
+    cachetools;python_version<"3.3"
+    enum34>=1.1;python_version<"3.4"
+    importlib_resources>=1.0;python_version<"3.7"
+    pathlib2>=2.3;python_version<"3.5"
+python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
+include_package_data = True
```
henryiii, line 53: This seems simpler than what was there before.

setup.cfg
```diff
@@ -1,14 +1,72 @@
...
+[options.extras_require]
+test =
+    pytest>=4.6
+
 [bdist_wheel]
 universal = 1
 
-[aliases]
-test=pytest
-
```
henryiii, line 6: Deprecated, users should not be calling `python setup.py <anything>`

setup.cfg
eduardo-rodrigues: Indeed. I look at that a while back and concluded at the time that Lark had changed behaviour significantly, in a way that almost looked like a bug. I can't remember the details now but then decided not to dwell on it immediate. I agree that we need to move to recent versions, also to potentially make things work for latest Python versions. I will create an issue ...

setup.cfg
eduardo-rodrigues: Fair.

setup.cfg
eduardo-rodrigues: Indeed. No need for a temporary variable. Thanks.

eduardo-rodrigues:
The issue with Python 3.5 is solvable for sure but I'm perfectly happy to drop support for it in the next minor release, so ...

henryiii:
Weird, why is lark missing on 3.5?

eduardo-rodrigues:
It isn't as I can see from the CI from my PR. So maybe a weird glitch? In any case it seems we can get rid of 3.5 anyway at this stage :-).

eduardo-rodrigues:
The CI for Python 3.5 in the step "Install package" says
```
Installing collected packages: UNKNOWN
  Running setup.py develop for UNKNOWN
Successfully installed UNKNOWN
Finishing: Install package
```
whereas for the other versions is starts collecting and installing the required packages.

setup.cfg
```diff
@@ -1,14 +1,72 @@
+[metadata]
+name = DecayLanguage
+description = A language to describe particle decays, and tools to work with them.
```
eduardo-rodrigues, line 3:
```suggestion
description = A language to describe, manipulate and convert particle decays.
```

eduardo-rodrigues:
OK, GA also fails on Python 3.5. So it's got to do with our setup. The only thing I see removed is `tests_require=extras["test"],`.

henryiii:
Ahh, pyproject.toml is missing. I think that’s the problem.

eduardo-rodrigues:
> Ahh, pyproject.toml is missing. I think that’s the problem.

Dunno. Will you take care of adding one?

eduardo-rodrigues (approved):
LGTM

(Noted that there is more on Scikit-HEP guidelines but this one PR is scoped and ready).

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[build-system]
+requires = [
+    "setuptools>=42",
+    "wheel",
+    "setuptools_scm[toml]>=3.4"
```
henryiii, line 5: We don't use `setuptools_scm` (yet).
```suggestion
```

henryiii:
Yes, one step at a time. If you want to add one of the other parts (setuptools_scm, more checks, etc), feel free. I'll try to ping you if I start working on one.

eduardo-rodrigues:
> Yes, one step at a time. If you want to add one of the other parts (setuptools_scm, more checks, etc), feel free. I'll try to ping you if I start working on one.

OK, sounds good. I actually found a bug in decay chain representations so will probably focus on that first, tough; for next week.

PR #131, opened 2021-03-29
DecayChain - minor fix, more tests and doc

eduardo-rodrigues:
For the record: the issue with `DecayChain.flatten()` is subtle in the sense that
```
dm1 = DecayMode(0.6770, "D0 pi+")
dm2 = DecayMode(0.0124, "K_S0 pi0")
dm3 = DecayMode(0.692, "pi+ pi-")
dm4 = DecayMode(0.98823, "gamma gamma")
dc = DecayChain("D*+", {"D*+": dm1, "D0": dm2, "K_S0": dm3, "pi0": dm4})
dc.flatten()
```
would result in the correct final state "gamma gamma pi+ pi+ pi-" but then if one closely looks at the daughters of that flattened decay mode one would see that `dc.flatten().decays[dc.mother].daughters == DaughtersDict(['gamma', 'gamma', 'pi+', 'pi+','pi-'])` would evaluate to `False`! That's because keys in the `Counter` instance (i.e. `DaughtersDict`) remained present with a count of 0. That's OK in general but makes no sense for a decay final state. The fix makes sure counts of 0 get removed.

The same "clean-up" is implemented for the constructor in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Version 0.11 (UNDER DEVELOPMENT)
 
 * Universal representation of decay chains:
+  - More documentation and testing.
+  - Allow default class method `DecayMode.from_pdgids()` mimicking the default constructor `DecayMode()`.
+  - Fix in logic of ``DecayChain.flatten()``.
   - `DecayChainViewer` class adapted to make use of ``graphviz``.
 * Dependencies:
   - Got rid of dependency on ``pydot``, as the package is no longer maintained.
--- a/decaylanguage/decay/decay.py
+++ b/decaylanguage/decay/decay.py
@@ -59,6 +59,8 @@ class DaughtersDict(Counter):
         if iterable and isinstance(iterable, str):
             iterable = iterable.split()
         super(DaughtersDict, self).__init__(iterable, **kwds)
+        # The "+" (see Counter's __pos__) makes sure that counts <=0 are removed!
+        +self
 
     def to_string(self):
         """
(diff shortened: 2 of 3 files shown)


eduardo-rodrigues:
Right, Python 2 and old Python 3 gets ever more annoying ... I will work on something that is more universal (to stay temporarily until we get rid of Py 3.5 and 2.7!).

henryiii:
My guess is something depends on dict ordering. The list of failing Pythons is exactly the list of unordered dict Pythons.

eduardo-rodrigues:
> My guess is something depends on dict ordering. The list of failing Pythons is exactly the list of unordered dict Pythons.

That's correct. But then I realised another issue in that the replacements for the flattening are not recursive/deep enough. I'm reworking the function.

henryiii:
Okay, I'm working on testing a bit. Producing too many warning for my liking :)

eduardo-rodrigues:
> Okay, I'm working on testing a bit. Producing too many warning for my liking :)

That's cool. Only 1-2 are real warnings that the tests look for.

henryiii:
Do you mind if I push to this branch? My fixes are based on this branch, shouldn't interfere with what you are doing.

henryiii:
I've reduced it to three warnings. Two are tricky because they show up in a matrix of tests. The other is a PyTest deprecation warning that is probably best left until after we drop Python 2 and can upgrade PyTest past version 4.

eduardo-rodrigues:
> Do you mind if I push to this branch? My fixes are based on this branch, shouldn't interfere with what you are doing.

Of course not. Go ahead.

eduardo-rodrigues:
Morning @henryiii, the PR is finally ready - had to scratch my head to come up with something better and correct!

eduardo-rodrigues:
I know you @henryiii are terribly busy. Am going to merge as this is a relevant fix.

BTW, I would like to make a 0.11.0 release imminently with this fix and the change of dependency to `graphviz`. We can then make a patch release with the improvements based on Scikit-HEP guidelines, as that does not impact the API/user. I won't do that before tomorrow so that you have time to react :-).

PR #133, opened 2021-03-31
style: developer guide

henryiii:
- chore: activate more developer guidelines
- style: pre-commit run -a for pyupgrade
- style: fix flake8 issues
- fix: correct all flake8 warnings

### DIFF
--- a/decaylanguage/dec/dec.py
+++ b/decaylanguage/dec/dec.py
@@ -42,6 +42,7 @@ from __future__ import absolute_import, division, print_function
 import os
 import warnings
 import re
+import sys
 import operator
 
 from six import StringIO
@@ -49,21 +50,17 @@ from six import StringIO
 from lark import Lark
 from lark import Tree, Transformer, Visitor
 
-from particle import Particle, ParticleNotFound
+from particle import Particle
 from particle.converters import PDG2EvtGenNameMap
-from particle.particle.enums import Charge, Charge_undo, Charge_mapping
 
 from .enums import PhotosEnum
 
 from ..utils import charge_conjugate_name
-from ..data import open_text
+
 from .. import data
 
-
 # New in Python 3
-try:
-    FileNotFoundError
-except NameError:
+if sys.version_info < (3,):
     FileNotFoundError = IOError
 
 
@@ -189,9 +186,9 @@ class DecFileParser(object):
         # Retrieve all info on the default Lark grammar and its default options,
         # effectively loading it
         opts = self.grammar_info()
-        extraopts = {
-            k: v for k, v in opts.items() if k not in ("lark_file", "parser", "lexer")
-        }
+        # extraopts = {
+        #     k: v for k, v in opts.items() if k not in ("lark_file", "parser", "lexer")
+        # }  not used later?
 
         # Instantiate the Lark parser according to chosen settings
         parser = Lark(self.grammar(), parser=opts["parser"], lexer=opts["lexer"])
@@ -694,11 +691,11 @@ All but the first occurrence will be discarded/removed ...""".format(
 
         dms = self._find_decay_modes(mother)
 
-        l = []
+        ls = []
         for dm in dms:
             if print_model:
                 dm_details = self._decay_mode_details(dm)
-                l.append(
+                ls.append(
                     (
                         dm_details[0],
                         (
@@ -714,14 +711,14 @@ All but the first occurrence will be discarded/removed ...""".format(
 
             else:
                 fsp_names = get_final_state_particle_names(dm)
-                l.append((get_branching_fraction(dm), "%-50s" % "  ".join(fsp_names)))
+                ls.append((get_branching_fraction(dm), "%-50s" % "  ".join(fsp_names)))
 
-        l.sort(key=operator.itemgetter(0), reverse=(not ascending))
+        ls.sort(key=operator.itemgetter(0), reverse=(not ascending))
 
-        for bf, info in l:
+        for bf, info in ls:
             print("{:12g} : {}".format(bf, info))
 
-    def build_decay_chains(self, mother, stable_particles=[]):
+    def build_decay_chains(self, mother, stable_particles=()):
         """
         Iteratively build the entire decay chains of a given mother particle,
         optionally considering, on the fly, certain particles as stable.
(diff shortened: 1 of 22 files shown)


decaylanguage/dec/dec.py
```diff
@@ -435,10 +432,10 @@ def _add_decays_to_be_copied(self):
                 copied_decay = match.__deepcopy__(None)
                 copied_decay.children[0].children[0].value = decay2copy
                 copied_decays.append(copied_decay)
-            except:
+            except Exception:
```
henryiii, line 435: No bare excepts! These might be able to be tightened up further, but at least now they don't catch memory errors, control C, etc.

decaylanguage/dec/dec.py
```diff
@@ -714,14 +711,14 @@ def print_decay_modes(
...
-        l.sort(key=operator.itemgetter(0), reverse=(not ascending))
+        ls.sort(key=operator.itemgetter(0), reverse=(not ascending))
 
-        for bf, info in l:
-            print("%12g : %s" % (bf, info))
+        for bf, info in ls:
+            print("{:12g} : {}".format(bf, info))
 
-    def build_decay_chains(self, mother, stable_particles=[]):
+    def build_decay_chains(self, mother, stable_particles=()):
```
henryiii, line 721: Never use a mutable structure in an argument default, it's highly bug-prone. (Note: many of these comments are probably to myself, as I wrote quite a few of theses...)

decaylanguage/dec/decparser.py
```diff
@@ -200,25 +195,25 @@ def addline(self, line):
...
-        except:
-            raise Exception("Cannot parse decay line: %s" % line)
-        ##        mod_found = False
-        ##         for x in known_models:
-        ##             if x in line:
-        ##                 mod_found = True
-        ##         if not mod_found:
-        ##             print '\n', line
+        except Exception:
+            raise RuntimeError("Cannot parse decay line: %s" % line)
```
henryiii, line 199: Always should be a bit more specific than "Exception" when throwing. RuntimeError is a pretty good "I don't know I want to throw" default.

tests/decay/test_decay.py
```diff
@@ -146,7 +146,7 @@ def test_DecayMode_charge_conjugate():
     assert "Decay model: TAUHADNU [-0.108, 0.775, 0.149, 1.364, 0.4]" in dm.describe()
 
     dd = DaughtersDict("pi- pi0 nu(tau)")
-    dd.charge_conjugate(pdg_name=True) == DaughtersDict("pi+ pi0 nu(tau)~")
+    assert dd.charge_conjugate(pdg_name=True) == DaughtersDict("pi+ pi0 nu(tau)~")
```
henryiii, line 149: This test tested nothing. I love this bugbear check :)

tests/dec/test_dec.py
```diff
@@ -187,14 +185,14 @@ def test_global_photos_flag():
     p = DecFileParser(DIR / "../data/defs-aliases-chargeconj.dec")
     p.parse()
 
-    assert p.global_photos_flag() == True
+    assert p.global_photos_flag()
```
henryiii, line 188: Never use `==` with None, True, or False.

decaylanguage/dec/decparser.py
eduardo-rodrigues: Yep.

henryiii:
After this one, I could move to a src/ structure, then we would pretty much be fully following the dev guidelines, I think. :)

tests/dec/test_dec.py
eduardo-rodrigues: Ah, did not know that. OK.

decaylanguage/dec/dec.py
```diff
@@ -189,9 +186,9 @@ def parse(self, include_ccdecays=True):
         # Retrieve all info on the default Lark grammar and its default options,
         # effectively loading it
         opts = self.grammar_info()
-        extraopts = {
-            k: v for k, v in opts.items() if k not in ("lark_file", "parser", "lexer")
-        }
+        # extraopts = {
+        #     k: v for k, v in opts.items() if k not in ("lark_file", "parser", "lexer")
+        # }  not used later?
```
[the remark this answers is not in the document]
eduardo-rodrigues: Right now I'm really lacking time so let me check this PR as a couple of things may need discussion ...

decaylanguage/dec/dec.py
eduardo-rodrigues: Oh, all these functions for building decay chains are my babies, actually, don't worry. I do tend to use lists as lot even when tuples are just fine, and this one is an example. Indeed the user is meant to give a "list" of particles they want to declare as stable for the purpose of playing with all the info parsed.

henryiii: As a good rule, lists, sets, and dicts should never be default arguments. Too easy to modify across calls. So swapping for tuple or frozenset is an easy fix. Dicts are harder, have to use None and check for it.

decaylanguage/dec/dec.py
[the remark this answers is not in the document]
eduardo-rodrigues: I guess you used flake8 to find this. Great. In fact it was meant to be used just below in the code when instantiating `Lark`:
`parser = Lark(self.grammar(), parser=opts["parser"], lexer=opts["lexer"])`
should be
`parser = Lark(self.grammar(), parser=opts["parser"], lexer=opts["lexer"], **extraopts)`.
Can you update within your PR?

decaylanguage/dec/dec.py
eduardo-rodrigues: Good catch. This was indeed overlooked.

eduardo-rodrigues (approved):
I left a tiny update to make. Otherwise great!

decaylanguage/dec/dec.py
```diff
@@ -191,10 +188,10 @@ def parse(self, include_ccdecays=True):
         opts = self.grammar_info()
         extraopts = {
             k: v for k, v in opts.items() if k not in ("lark_file", "parser", "lexer")
-        }
+        }  not used later?
```
eduardo-rodrigues, line 191:
```suggestion
        }
```

PR #137, opened 2021-04-01
chore: move to src dir format

henryiii:
Exposes some issues using mypy - disabled the mypy check for now, will renable and fix the issues in the next PR (shouldn't rename and change files in the same commit).


eduardo-rodrigues (approved):
LGTM. Thanks.

PR #139, opened 2021-04-01
fix: add mypy and fix/hide issues

henryiii:
This just does a basic "untyped" mypy check, and still managed to find a few issues. I'll comment on a few of the changes, but in general, mostly just masked them out for now, with 1-3 real fixes.

### DIFF
--- a/src/decaylanguage/data/__init__.py
+++ b/src/decaylanguage/data/__init__.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
-try:
-    from importlib.resources import open_text
-except ImportError:
+import sys
+
+if sys.version_info < (3, 7):
     from importlib_resources import open_text
+else:
+    from importlib.resources import open_text
 
 __all__ = ("open_text",)
--- a/src/decaylanguage/dec/dec.py
+++ b/src/decaylanguage/dec/dec.py
@@ -131,7 +131,7 @@ class DecFileParser(object):
             # Content of input file(s)
             self._dec_file = stream.read()
         else:
-            self._dec_file_names = None  # type: ignore
+            self._dec_file_names = []  # type: list
             self._dec_file = None  # type: ignore
 
         self._parsed_dec_file = None  # Parsed decay file
--- a/src/decaylanguage/dec/dec.py
+++ b/src/decaylanguage/dec/dec.py
@@ -798,8 +798,7 @@ All but the first occurrence will be discarded/removed ...""".format(
 
             info.append(d)
 
-        retval = {mother: info}
-        return retval
+        return {mother: info}
 
     def __repr__(self):
         if self._parsed_dec_file is not None:
--- a/src/decaylanguage/dec/dec.py
+++ b/src/decaylanguage/dec/dec.py
@@ -131,8 +131,8 @@ class DecFileParser(object):
             # Content of input file(s)
             self._dec_file = stream.read()
         else:
-            self._dec_file_names = None
-            self._dec_file = None
+            self._dec_file_names = None  # type: ignore
+            self._dec_file = None  # type: ignore
 
         self._parsed_dec_file = None  # Parsed decay file
         self._parsed_decays = None  # Particle decays found in the decay file
@@ -154,7 +154,7 @@ class DecFileParser(object):
         stream.seek(0)
 
         _cls = cls()
-        _cls._dec_file_names = "<dec file input as a string>"
+        _cls._dec_file_names = ["<dec file input as a string>"]
         _cls._dec_file = stream.read()
 
         return _cls
@@ -422,7 +422,7 @@ class DecFileParser(object):
         # match name -> position in list self._parsed_decays
         name2treepos = {
             t.children[0].children[0].value: i
-            for i, t in enumerate(self._parsed_decays)
+            for i, t in enumerate(self._parsed_decays)  # type: ignore
         }
 
         # Make the copies taking care to change the name of the mother particle
@@ -430,7 +430,7 @@ class DecFileParser(object):
         misses = []
         for decay2copy, decay2becopied in decays2copy.items():
             try:
-                match = self._parsed_decays[name2treepos[decay2becopied]]
+                match = self._parsed_decays[name2treepos[decay2becopied]]  # type: ignore
                 copied_decay = match.__deepcopy__(None)
                 copied_decay.children[0].children[0].value = decay2copy
                 copied_decays.append(copied_decay)
@@ -445,7 +445,7 @@ Skipping creation of these copied decay trees.""".format(
             warnings.warn(msg)
 
         # Actually add all these copied decays to the list of decays!
-        self._parsed_decays.extend(copied_decays)
+        self._parsed_decays.extend(copied_decays)  # type: ignore
 
     def _add_charge_conjugate_decays(self):
         """
@@ -477,7 +477,7 @@ Skipping creation of these copied decay trees.""".format(
         # Cross-check - make sure charge conjugate decays are not defined
         # with both 'Decay' and 'CDecay' statements!
         mother_names_decays = [
-            get_decay_mother_name(tree) for tree in self._parsed_decays
+            get_decay_mother_name(tree) for tree in self._parsed_decays  # type: ignore
         ]
 
         duplicates = [n for n in mother_names_ccdecays if n in mother_names_decays]
@@ -508,7 +508,7 @@ The 'CDecay' definition(s) will be ignored ...""".format(
         # match name -> position in list self._parsed_decays
         name2treepos = {
             t.children[0].children[0].value: i
-            for i, t in enumerate(self._parsed_decays)
+            for i, t in enumerate(self._parsed_decays)  # type: ignore
         }
 
         trees_to_conjugate = []
@@ -516,7 +516,7 @@ The 'CDecay' definition(s) will be ignored ...""".format(
         for ccname in mother_names_ccdecays:
             name = find_charge_conjugate_match(ccname, dict_cc_names)
             try:
-                match = self._parsed_decays[name2treepos[name]]
+                match = self._parsed_decays[name2treepos[name]]  # type: ignore
                 trees_to_conjugate.append(match)
             except Exception:
                 misses.append(ccname)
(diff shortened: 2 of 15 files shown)


src/decaylanguage/data/__init__.py
```diff
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
-try:
-    from importlib.resources import open_text
-except ImportError:
+import sys
+
+if sys.version_info < (3, 7):
```
henryiii, line 4: It's better to compare with sys.version_info. It:
* Tells a reader exactly where one works vs. the other
* Works with static tools, like mypy
* Can support using a backport after it was introduced (if we use the newer 3.8/3.9  API of importlib.resouces, for example).

src/decaylanguage/dec/dec.py
```diff
@@ -131,8 +131,8 @@ def __init__(self, *filenames):
             # Content of input file(s)
             self._dec_file = stream.read()
         else:
-            self._dec_file_names = None
-            self._dec_file = None
+            self._dec_file_names = None  # type: ignore
+            self._dec_file = None  # type: ignore
```
henryiii, line 135: Don't normally do this, but it was faster/easier for me not to look up or guess the types. We should type particle first.

src/decaylanguage/dec/dec.py
```diff
@@ -154,7 +154,7 @@ def from_string(cls, filecontent):
         stream.seek(0)
 
         _cls = cls()
-        _cls._dec_file_names = "<dec file input as a string>"
+        _cls._dec_file_names = ["<dec file input as a string>"]
```
henryiii, line 157: Types should be consistent if possible; this is a list normally.

src/decaylanguage/dec/dec.py
```diff
@@ -422,15 +422,15 @@ def _add_decays_to_be_copied(self):
         # match name -> position in list self._parsed_decays
         name2treepos = {
             t.children[0].children[0].value: i
-            for i, t in enumerate(self._parsed_decays)
+            for i, t in enumerate(self._parsed_decays)  # type: ignore
```
henryiii, line 425: This (and _most_ of the ignores below) are due to the fact that a multi-state class is a bad design (I've written many of them). Basically, this class is only valid if you call `.parse`; so every method should have a `assert self._parsed_decays is not None` at the front, or something similar. It's better if there were two classes (parsed/unparsed) or parse was part of the constructor, etc.

src/decaylanguage/dec/dec.py
```diff
@@ -570,7 +570,7 @@ def _check_parsed_decays(self):
         """
         # Issue a helpful warning if duplicates are found
         lmn = self.list_decay_mother_names()
-        duplicates = []
+        duplicates = set()  # type: set
```
henryiii, line 573: Don't do this normally; the reason it's complaining is it wants to know what the type is in the set. `Set[<something>]` would be the proper fix, but I don't want to bother with importing typing and figuring out what is in each one of these, so I'm cheating for now by forcing it to the generic type here.

src/decaylanguage/dec/dec.py
```diff
@@ -798,8 +798,8 @@ def build_decay_chains(self, mother, stable_particles=()):
 
             info.append(d)
 
-        info = {mother: info}
-        return info
+        retval = {mother: info}
```
henryiii, line 801: Best not to reuse variables, especially when changing types.

src/decaylanguage/dec/decparser.py
```diff
@@ -223,25 +220,22 @@ def addline(self, line):
         if killindex is None:
             print(spl)
             raise Exception("No decay model specified: %s" % line)
-        #        print spl[killindex:]
         decay = Decay()
         decay.bf = bf
-        #        print self.current_decay_top.decay_of, 'to',
         for part in spl[1:killindex]:
-            #            print part,
```
henryiii, line 231: This is git, no commented out code please, unless there's a really good reason. Can always pull it from history.

It might be a great idea to add `logging` support, though, so you could enable the logger and see DEBUG or TRACE output from things like this. :)

src/decaylanguage/decay/viewer.py
```diff
@@ -12,16 +12,13 @@
...
-except AttributeError:
-    counter = itertools.count().next
-
 import graphviz
 
 from particle import latex_to_html_name
 from particle.converters.bimap import DirectionalMaps
 
+counter = iter(itertools.count())
+
```
henryiii, line 21: Use the `next()` builtin instead, that's calls the right thing on both Pythons. Also call `iter` if you plan to manually iterate something.

src/decaylanguage/modeling/__init__.py
```diff
@@ -2,4 +2,4 @@
 from .amplitudechain import LS
 from .amplitudechain import AmplitudeChain
 
-__all__ = (LS, AmplitudeChain)
+__all__ = ("LS", "AmplitudeChain")
```
henryiii, line 5: These should always be strings.

src/decaylanguage/modeling/ampgen2goofit.py
```diff
@@ -31,7 +31,7 @@ def ampgen2goofit(filename, ret_output=False):
         output = StringIO()
         printer = partial(print, file=output)
     else:
-        printer = print
+        printer = print  # type: ignore
```
henryiii, line 34: Proper fix is to predeclare this, but that's Python 3.6 syntax. The other fix would be to add a type on both statements. Or just ignore it.

src/decaylanguage/utils/errors.py
```diff
@@ -10,7 +10,5 @@
 
 
 class LineFailure(RuntimeError):
-    def __init__(self, line, message, *args, **kwargs):
-        super(LineFailure, self).__init__(
-            "{}: {}".format(line, message), *args, **kwargs
-        )
+    def __init__(self, line, message):
+        super(LineFailure, self).__init__("{}: {}".format(line, message))
```
henryiii, line 14: RuntimeError doesn't actually have any other args, so MyPy complains. 50/50 on having it check this, sometimes it's wrong, like with Mixins. But here it's fine.

src/decaylanguage/utils/utilities.py
```diff
@@ -53,7 +53,7 @@ def filter_lines(matcher, inp):
     Filter out lines into new variable if they match a regular expression
     """
     output = [
-        matcher.match(ln).groupdict() for ln in input if matcher.match(ln) is not None
+        matcher.match(ln).groupdict() for ln in inp if matcher.match(ln) is not None
     ]
-    new_inp = [ln for ln in input if matcher.match(ln) is None]
+    new_inp = [ln for ln in inp if matcher.match(ln) is None]
```
henryiii, line 58: This is a bug, probably introduced by me recently.

src/decaylanguage/dec/dec.py
```diff
@@ -798,8 +798,8 @@ def build_decay_chains(self, mother, stable_particles=()):
...
+        return retval
```
henryiii, line 802:
```suggestion
        return {mother: info}
```

henryiii: Found by Sourcery.ai

src/decaylanguage/utils/utilities.py
eduardo-rodrigues: Could have nasty effects! Nice catch.

src/decaylanguage/modeling/__init__.py
eduardo-rodrigues: Your baby, did not touch that :D.

henryiii: I'm likely the author of many/most of the issues, just pointing things out for instructional purposes in the future for all of us. :)

src/decaylanguage/dec/decparser.py
eduardo-rodrigues: In fact this file is to be removed as it was what we got from Peter Onyisi in the old times. It is not quite working and ... But I will act on that soon.

src/decaylanguage/modeling/__init__.py
eduardo-rodrigues: Oh, most probably me writing non-optimal code overall. I bet that ;-). But yes, your comments are very much appreciated for exactly the reason you say. Thank you for that.

src/decaylanguage/data/__init__.py
eduardo-rodrigues: Makes sense :-).

src/decaylanguage/dec/dec.py
```diff
@@ -131,8 +131,8 @@ def __init__(self, *filenames):
             # Content of input file(s)
             self._dec_file = stream.read()
         else:
-            self._dec_file_names = None
-            self._dec_file = None
+            self._dec_file_names = None  # type: ignore
```
eduardo-rodrigues, line 134:
```suggestion
            self._dec_file_names = []
```

eduardo-rodrigues: This one is easy, so let me help you trivially.

src/decaylanguage/dec/dec.py
eduardo-rodrigues: Indeed I made a mistake. Should be a list.

src/decaylanguage/dec/dec.py
henryiii: It will want to know what's inside the list, so quieting it with `type: list`. Ideally it should be `type: List[str]` I expect, but don't want to bother importing things from typing just yet.

henryiii: Ahh, you were right, this wasn't the definition. 🤦

src/decaylanguage/dec/dec.py
```diff
@@ -131,8 +131,8 @@ def __init__(self, *filenames):
             # Content of input file(s)
             self._dec_file = stream.read()
         else:
-            self._dec_file_names = None
-            self._dec_file = None
+            self._dec_file_names = []  # type: list
```
henryiii, line 134:
```suggestion
            self._dec_file_names = []
```

src/decaylanguage/dec/dec.py
eduardo-rodrigues: I did a check on the parsing in several places but for sure not everywhere. This is something we can revisit. Thanks.

src/decaylanguage/dec/dec.py
eduardo-rodrigues: And indeed I use a set later one, hence your change is logical.

eduardo-rodrigues (approved):
Thank you for these design improvements!

PR #150, opened 2021-05-20
ci: normalize dist following guidelines

henryiii:


eduardo-rodrigues:
Thanks for this @henryiii.

As for `Particle` also here we should get rid of the redundant Azure pipelines and restored coverage testing within GAs.

eduardo-rodrigues (approved):

PR #163, opened 2021-06-22
chore: bump mypy and move to toml config

henryiii:
@eduardo-rodrigues I've updated the config - and this one mypy error is left:

```
src/decaylanguage/dec/decparser.py:92: error: "Type[Particle]" has no attribute "from_search_list"
```

Is this correct, is that something we changed?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,11 +39,12 @@ repos:
     stages: [manual]
 
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v0.812
+  rev: v0.902
   hooks:
   - id: mypy
     files: src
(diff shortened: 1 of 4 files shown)


eduardo-rodrigues:
> @eduardo-rodrigues I've updated the config - and this one mypy error is left:
> 
> ```
> src/decaylanguage/dec/decparser.py:92: error: "Type[Particle]" has no attribute "from_search_list"
> ```
> 
> Is this correct, is that something we changed?

Sorry, only now replying. Yes, this is correct. That method got superseded and the issue never one in reality because decparser is not used and not fully working. It's been on my to-dos to get back to it but low priority ... I will try and come back to work on the file ...

eduardo-rodrigues (approved):
LGTM!

Issue #156, opened 2021-06-07
CI clean-up + enhancement needed


eduardo-rodrigues:
- [x] Remove Azure pipelines and rely fully on Github Actions ([PR 177](https://github.com/scikit-hep/decaylanguage/pull/177)).
- [ ] Make sure we continue to test on Linux, masOS and Windows as previously on Azure  ([PR 179](https://github.com/scikit-hep/decaylanguage/pull/179)).
- [x] Make coverage tests live again ([PR 178](https://github.com/scikit-hep/decaylanguage/pull/178)).

PR #179, opened 2021-08-18
Run the CI on Linux, MacOS and Windows

eduardo-rodrigues:
Closes https://github.com/scikit-hep/decaylanguage/issues/156.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,17 +13,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest]
         python-version: ['2.7', '3.5', '3.6', '3.7', '3.8', '3.9']
+        include:
+          - {os: macos-latest, python-version: '2.7'}
+          - {os: macos-latest, python-version: '3.9'}
+          - {os: windows-latest, python-version: '3.7'}
+          - {os: windows-latest, python-version: '3.9'}
         exclude:
-          - os: macos-latest
-            python-version: '3.7'
-          - os: macos-latest
-            python-version: '3.8'
-          - os: windows-latest
-            python-version: '3.7'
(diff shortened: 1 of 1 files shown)


eduardo-rodrigues:
Hi @henryiii, this is now ready for review. The CI now really checks a fair matrix of OSes and Python versions. This gives confidence on what is supported :-).

.github/workflows/ci.yml
```diff
@@ -8,20 +8,26 @@ on:
...
+        python-version: ['2.7', '3.5', '3.6', '3.7', '3.8', '3.9']
+        exclude:
+          - os: macos-latest
+            python-version: '3.7'
+          - os: macos-latest
+            python-version: '3.8'
+          - os: windows-latest
+            python-version: '3.7'
+          - os: windows-latest
+            python-version: '3.8'
```
henryiii, line 26: This is a very large matrix, and is about to get bigger if we add Python 3.10 (now if we kill python 2 and 3.5... ;) ) MacOS builds are limited to 5x at a time, and limits are shared across an org, I believe. How about just adding a few, rather than excluding a few? Also, no one uses Python 2.7 on Windows, most binaries packages for it are frozen due to Microsoft pulling the 2008 compiler download. For pure-Python things like this, it's pretty unlikely to have major differences.

```suggestion
        os: [ubuntu-latest]
        python-version: ['2.7', '3.5', '3.6', '3.7', '3.8', '3.9']
        include:
          - {os: macos-latest, python-version: '2.7'}
          - {os: macos-latest, python-version: '3.9'}
          - {os: windows-latest, python-version: '3.7'}
          - {os: windows-latest, python-version: '3.9'}
        exclude:
          - {os: ubuntu-latest, python-version: '3.7'}
```

eduardo-rodrigues: Right, one would never continue adding without the removal of some. That's only fair. But agreed that the matrix can be made fairly smaller.

PR #186, opened 2021-09-02
Prepare release 0.12.0

eduardo-rodrigues:


setup.cfg
```diff
@@ -4,8 +4,8 @@ description = A language to describe, manipulate and convert particle decays
 long_description = file: README.md
 long_description_content_type = text/markdown
 url = https://github.com/scikit-hep/decaylanguage
-author = Henry Fredrick Schreiner III, Eduardo Rodrigues
-author_email = henry.schreiner@cern.ch, eduardo.rodrigues@cern.ch
+author = Eduardo Rodrigues, Henry Fredrick Schreiner III
```
eduardo-rodrigues, line 7: I changed this to be in a more natural order, also because most of the package is actually the stuff I put in by now. I hope this is OK with you.

henryiii:
Why did you remove Python 3.5 support? I don't see any benefit at all in removing it if we still support 2.7. There's no cleanup / new syntax that you can use if we still support 2.7. I'd love to drop 2.7 from DecayLanguage, personally. :) I think we should drop it on all remaining projects Jan 1, 2022.

setup.cfg
henryiii: That's fine. :)

eduardo-rodrigues:
Oh, I could have commented on that too, fair enough. In fact the CI failed because Particle does not support Py 3.5 since a while. I had forgotten that. No worries, I don't think that is going to be an issue for anyone these days ... even big experiments such as LHCb are now on Python 3.7 and moving up from there.

I agree with you, BTW, that we can drop support for Python 2.7 next year. It is easy to say "just in case some stacks ..." but one does need to draw a line somewhere.

henryiii:
I'm just not a fan of dropping a version with no benefit at all. When I drop 2.7 / 3.5, it comes with a massive simplification (boost-histogram 1.0 literally has fewer lines of code than 0.12, but more features); it's clearly a maintenance win. For a built package, you also get lower PyPI storage usage and a big CI improvement. Dropping 3.5 by itself generally does nothing at all, no simplification, it's just an arbitrary choice to be more strict on users for no benefit to us. Why did Particle drop it? Particle still should support 2.7, so I'd assume there's no benefit there either?

henryiii:
(For the record, though, I don't think supporting 3.5 is very necessary, I'd just like to drop both (3.5, 2.7) rather than dropping one).

henryiii:
(I'd also add Python 3.10 to the tests and to the supported versions when dropping a version)

henryiii:
Haha, https://github.com/scikit-hep/particle/pull/323/files might have been a mistake on my part due to an update with pyupgrade considering 3.6+ to be the default (combined with the fact we weren't testing it)

PR #187, opened 2021-09-03
chore: use newer importlib resources syntax

### LINKED ISSUE #138: Update importlib.resources usage
We should use the newer interface (3.8+ or 3.9+ with backport), it's much simpler to use. It's also very unevenly used in the current package; hopefully we can use it consistently once we use the new API.

henryiii:
Closes #138

Can also check into updating a few raw string usages into basepath usages, might be a followup.

- chore: add nox support
- feat: use newer importlib resources


eduardo-rodrigues:
Thank you for this - I had added the work to my to-do list but since you got to it :-).

PR #188, opened 2021-09-03
chore: update and fix warnings

### LINKED ISSUE #130: Remove constraint on Lark version and update to most recent ones
See discussion at https://github.com/scikit-hep/decaylanguage/pull/129#discussion_r601668378.

henryiii:
This fixes a few things, corrects the warnings showing up in the tests, and also bumps lark because it doesn't seem to break anything to do so? Maybe it was a bug that was fixed? Closes #130

### DIFF
--- a/src/decaylanguage/dec/dec.py
+++ b/src/decaylanguage/dec/dec.py
@@ -277,7 +277,8 @@ class DecFileParser(object):
             # Conversion to handle pathlib on Python < 3.6:
             filename = str(filename)
 
-            self._grammar = open(filename).read()
+            with open(filename) as f:
+                self._grammar = f.read()
 
         self._grammar_info = dict(
             lark_file=filename, parser=parser, lexer=lexer, **options
--- a/src/decaylanguage/modeling/ampgentransform.py
+++ b/src/decaylanguage/modeling/ampgentransform.py
@@ -4,6 +4,9 @@
 # Distributed under the 3-clause BSD license, see accompanying file LICENSE
 # or https://github.com/scikit-hep/decaylanguage for details.
 
+from collections import OrderedDict
+from typing import Any, Dict
+
 from lark import Transformer, Tree
 
 
@@ -42,7 +45,9 @@ class AmpGenTransformer(Transformer):
 
     def decay(self, lines):
         (particle,) = lines[0].children
-        dic = {"name": str(particle), "daughters": []}
+        dic = OrderedDict()  # type: Dict[str, Any]
+        dic["name"] = str(particle)
+        dic["daughters"] = []
 
         for line in lines[1:]:
             if line.data == "subdecay":
--- a/tests/decay/test_viewer.py
+++ b/tests/decay/test_viewer.py
@@ -49,18 +49,18 @@ def test_simple_decay_chain():
 
 
 checklist_decfiles = (
-    (DIR / "../data/test_Bc2BsPi_Bs2KK.dec", "B_c+sig"),
-    (DIR / "../data/test_Bd2DDst_Ds2DmPi0.dec", "B0sig"),
-    (DIR / "../data/test_Bd2DmTauNu_Dm23PiPi0_Tau2MuNu.dec", "B0sig"),
-    (DIR / "../data/test_Bd2DMuNu.dec", "anti-B0sig"),
-    (DIR / "../data/test_Bd2DstDst.dec", "anti-B0sig"),
-    (DIR / "../data/test_example_Dst.dec", "D*+"),
-    (DIR / "../data/test_Xicc2XicPiPi.dec", "Xi_cc+sig"),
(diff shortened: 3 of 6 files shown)


src/decaylanguage/dec/dec.py
```diff
@@ -277,7 +277,8 @@ def load_grammar(self, filename=None, parser="lalr", lexer="standard", **options
             # Conversion to handle pathlib on Python < 3.6:
             filename = str(filename)
 
-            self._grammar = open(filename).read()
+            with open(filename) as f:
+                self._grammar = f.read()
```
henryiii, line 281: This is a bug, the file was not being closed. Using `-Werror` causes this bug to show up.

src/decaylanguage/modeling/ampgentransform.py
```diff
@@ -42,7 +45,9 @@ def cplx_decay_line(self, lines):
 
     def decay(self, lines):
         (particle,) = lines[0].children
-        dic = {"name": str(particle), "daughters": []}
+        dic = OrderedDict()  # type: Dict[str, Any]
+        dic["name"] = str(particle)
+        dic["daughters"] = []
```
henryiii, line 50: This was the thing that broke when upgrading lark; it was reversing this item, and dicts are not reversible until about 3.7 or 3.8.

src/decaylanguage/dec/dec.py
eduardo-rodrigues: Indeed :S.

For my interest in which comment did you run -Werror? That's not something I've been playing with, it seems.

src/decaylanguage/modeling/ampgentransform.py
eduardo-rodrigues: Ah, but I did not touch that part of the code and I would bet that when I tried to upgrade Lark a while back (certainly over a year ago) the issues were with my decay file parser. But if things are solved for the whole package, then cool. It is not impossible that Lark itself had an issue at some point in history ... Excellent news 👍!

eduardo-rodrigues:
> This fixes a few things, corrects the warnings showing up in the tests, and also bumps lark because it doesn't seem to break anything to do so? Maybe it was a bug that was fixed? Closes #130

It seems so, yes, because Lark was giving plenty of errors in our tests when I tried an upgrade over a year ago! Oh wow.

src/decaylanguage/dec/dec.py
henryiii: I ran `nox -s tests -- -Werror`; if you are directly running pytest, you can run `pytest -Werror`. I considered making it default, or adding it to the GitHub Actions run, but that would turn all warnings into errors, and if there was a warning for something else we don't control, we might not want to break because of it.

eduardo-rodrigues: Nice! I should check more carefully all the pytest options. Thanks for the tip.

eduardo-rodrigues: Yes, better not make it default for the reason you mention. Hopefully warnings won't come up too often.

henryiii: `--strict` can't be used without generating a warning, as it's now `--strict-markers` (which is much clearer, it actually tells you what it's being strict about), but we can't use that until we drop Python 2, since the last PyTest to support Python 2 didn't have `--strict-markers` yet. That's one example (though I just removed it for now, so it technically does pass with `-Werror` at the moment).

tests/decay/test_viewer.py
```diff
@@ -49,26 +49,30 @@ def test_simple_decay_chain():
...
     """
     This test effectively checks whether any box node (node with subdecays)
     gets more than one arrow to it, which would show a bug
     in the creation of the DOT file recursively parsing the built decay chain.
     """
     p = DecFileParser(decfilepath, DIR / "../../src/decaylanguage/data/DECAY_LHCB.DEC")
 
-    p.parse()
+    if dup:
+        with pytest.warns(UserWarning, match="pi0"):
```
eduardo-rodrigues, line 72: What's the reason for this new warning showing up? Something to fix upstream?

eduardo-rodrigues (approved):
Just a little question otherwise all great. Excellent that the issue with Lark is gone!

tests/decay/test_viewer.py
henryiii: It's an old warning, we just weren't doing anything with it before. It's a duplicate particle warning, and I think it was intentional in these examples, they were supposed to have a duplicate particle definition warning?

henryiii: (I've been meaning to handle it for a while)

eduardo-rodrigues: Oh, now I remember - yes, I did add some checks of that kind because the parser is supposed to run for checks for silly user mistakes :D. Thanks as ever.

PR #195, opened 2021-10-05
Fix B904 warnings from pre-commit

eduardo-rodrigues:
Follow-up from https://github.com/scikit-hep/decaylanguage/pull/194.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 
 repos:
 - repo: https://github.com/asottile/pyupgrade
-  rev: v2.28.0
+  rev: v2.29.0
   hooks:
   - id: pyupgrade
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -78,5 +78,5 @@ addopts =
 
 [flake8]
 max-complexity = 24
-ignore = E203, E231, E501, E722, W503, B950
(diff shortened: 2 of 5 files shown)


eduardo-rodrigues:
OK, was too ambitious/quick and the update is Python 3 only. Closing.

henryiii:
We need to disable B904 until Py2 is dropped. Let's drop Py2 Jan 1, 2022 in decaylanguage and particle. (Though sooner than that is fine with me for DecayLanguage, I very much doubt it's important to keep Py2 for such a specialized package.

eduardo-rodrigues:
> We need to disable B904 until Py2 is dropped. Let's drop Py2 Jan 1, 2022 in decaylanguage and particle. (Though sooner than that is fine with me for DecayLanguage, I very much doubt it's important to keep Py2 for such a specialized package.)

Right, we already agreed on dropping Py2 early next year. But I also agree with the fact that I would doubt anyone using DecayLanguage would do it in a Py2 environment, hence fine with dropping Py2 now. That would make it easier with the issue at hand since temporarily disabling B904 for now is likely to be forgotten when we would drop Py2 in a few months. In short, happy to drop Py2 now and upgrade pyupgrade as in https://github.com/scikit-hep/decaylanguage/pull/194.

PR #210, opened 2022-01-05
Update copyright to 2022

eduardo-rodrigues:


henryiii:
FYI: https://hynek.me/til/copyright-years/

eduardo-rodrigues:
> FYI: https://hynek.me/til/copyright-years/

Thanks for the link. I confess that anyway I do prefer to give a sense of having licenses updated once a year with the present year explicit. It is tiny work compared to normal work :-).

Issue #211, opened 2022-01-06
_repr_svg_() is deprecated from graphviz v0.19


vvsagar:
It looks like `graphviz.Digraph`'s `_repr_svg_()` is replaced by `_repr_mimebundle_(include, exclude)` now: see [changelog](https://graphviz.readthedocs.io/en/stable/changelog.html#version-0-19).

This seems to lead to `AttributeError` when using `DecayChainViewer` in notebooks.

I am still figuring out how to get the rendered image using `_repr_mimebundle_()` instead of just SVG string.

vvsagar:
Okay, using `_repr_image_svg_xml()` seems to do the trick.

eduardo-rodrigues:
Thank you for reporting. We will have to see whether we pin 0.18.* for now or make the code work with both 0.18 and 0.19 ...

Out of curiosity, what work are you using DecayLanguage for?

vvsagar:
Or you could also force >=0.19, right?

I am trying to understand the B-tagging at Belle (II) and using DecayLanguage to study the current MC modeling, to see how well it agrees with PDG, and how it can be further improved. Along the way I made some custom functions like `build_decay_chains_to_inclusive_fsp`, `build_decay_chains_to_specific_fsp` (which identifies all possible intermediate resonances), modified `DecayChainViewer` to show effective BF at fsp level and `DecayChainToTable`.
You can see some of it in [this example notebook](https://github.com/vvsagar/decaylanguage/blob/inclusive-bf-studies/notebooks/ExampleInclusiveBF.ipynb). I was mostly hacking things for my convinence, but if you think any of this would be useful features, I can optimize and make official PRs.

eduardo-rodrigues:
> Or you could also force >=0.19, right?

This would not be a good idea just now since version 0.19 is very new. Users should still have the choice.

[...]

PR #213, opened 2022-01-14
Adapt to work with graphviz >= 0.19 that broke parts of the API

eduardo-rodrigues:
Fixes https://github.com/scikit-hep/decaylanguage/issues/211.

### DIFF
--- a/src/decaylanguage/decay/viewer.py
+++ b/src/decaylanguage/decay/viewer.py
@@ -240,4 +240,7 @@ class DecayChainViewer:
         """
         IPython display in SVG format.
         """
-        return self._graph._repr_svg_()
+        try:
+            return self._graph._repr_image_svg_xml()  # for graphviz >= 0.19
+        except AttributeError:
+            return self._graph._repr_svg_()  # for graphviz < 0.19
--- a/src/decaylanguage/decay/viewer.py
+++ b/src/decaylanguage/decay/viewer.py
@@ -236,11 +236,11 @@ class DecayChainViewer:
     def _get_edge_defaults(self):
         return dict(fontcolor="#4c4c4c", fontsize="11")
 
-    def _repr_svg_(self):
+    def _repr_mimebundle_(self, include=None, exclude=None, **kwargs):
         """
-        IPython display in SVG format.
+        IPython display helper.
         """
         try:
-            return self._graph._repr_image_svg_xml()  # for graphviz >= 0.19
+            return self._graph._repr_mimebundle_(include=include, exclude=exclude, **kwargs)
         except AttributeError:
-            return self._graph._repr_svg_()  # for graphviz < 0.19
+            return {"image/svg+xml": self._graph._repr_svg_()}  # for graphviz < 0.19
--- a/src/decaylanguage/modeling/decay.py
+++ b/src/decaylanguage/modeling/decay.py
@@ -121,4 +121,7 @@ class ModelDecay:
             return d
 
         def _repr_svg_(self):
-            return self._make_graphviz()._repr_svg_()
+            try:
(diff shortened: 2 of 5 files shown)


src/decaylanguage/decay/viewer.py
```diff
@@ -240,4 +240,7 @@ def _repr_svg_(self):
         """
         IPython display in SVG format.
         """
-        return self._graph._repr_svg_()
+        try:
+            return self._graph._repr_image_svg_xml()  # for graphviz >= 0.19
+        except AttributeError:
+            return self._graph._repr_svg_()  # for graphviz < 0.19
```
henryiii, line 246: This is exchanging a protocol for a private API. Instead we should use 
https://github.com/xflr6/graphviz/blob/723cf0213a112b458f3d5023f46daf0f2961ffa5/graphviz/jupyter_integration.py#L59

We should change over to providing `_repr_mimebundle_` instead, then build the mime bundle from `_repr_svg_()` if the `_graph` does not have the matching method.

henryiii: (And probably throw an error if the requested mime bundle is not svg and graphviz is old)

eduardo-rodrigues: Fair enough - this works but is not the most elegant.

You mean replace the `def _repr_svg_(self):` method with
```
    def _repr_mimebundle_(self, include=None, exclude=None, **kwargs):
        """
        IPython display in SVG format.
        """
        try:
            return self._graph._repr_mimebundle_(include=include, exclude=exclude, **kwargs)
        except AttributeError:
            return self._graph._repr_svg_()  # for graphviz < 0.19
```
?
This does not need to throw an exception since SVG will be picked up given the default `include=None` and will work with old graphviz versions, which I think is a must since 0.19 is very new.

henryiii: I think it's `return {"image/svg+xml": self._graph._repr_svg_()} `. I just meant it should throw an error if you request something other than svg, since this interface lets you request png and jpeg. But it's probably fine, it can still tell what got returned.

eduardo-rodrigues: Yep, was looking at that right now since the 2nd return is no longer a dict as expected :-). Let me test just in case ...

eduardo-rodrigues: OK

src/decaylanguage/modeling/decay.py
```diff
@@ -121,4 +121,9 @@ def _make_graphviz(self):
             return d
 
         def _repr_svg_(self):
-            return self._make_graphviz()._repr_svg_()
+            try:
+                return (
+                    self._make_graphviz()._repr_image_svg_xml()
+                )  # for graphviz >= 0.19
+            except AttributeError:
+                return self._make_graphviz()._repr_svg_()  # for graphviz < 0.19
```
henryiii, line 129: Same thing here! Otherwise looks good.

vvsagar:
Would it be useful/necessary to write a test function for this as well?

src/decaylanguage/modeling/decay.py
eduardo-rodrigues: Ah, last night I somehow forgot to push the changes here! Done now.

eduardo-rodrigues:
> Would it be useful/necessary to write a test function for this as well?

Tests are always useful/necessary. The old way has been tested for ever and the new one works. But looking carefully I see that the demo notebook is not executed in the CI as we do in the `Particle` package, for example. I will implement this as a follow-up ... unless you fancy a contribution to the package and want to deal with this? Let me know @vvsagar.

PR #216, opened 2022-01-18
Introduce type checking

eduardo-rodrigues:
This first "stab" does not implement type checking in the `modeling` submodule. That's for a follow-up.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,11 @@ write_to = "src/decaylanguage/_version.py"
 profile = "black"
 
 [tool.mypy]
+files = [
+    "src/decaylanguage/decay",
+    "src/decaylanguage/dec",
+    "src/decaylanguage/utils",
+]
 exclude = "src/decaylanguage/(data|modeling)"
 python_version = "3.9"
 warn_unused_configs = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ profile = "black"
 [tool.mypy]
 files = "src"
 exclude = [
-    'decaylanguage\.modeling\.*',
+    '^decaylanguage/modeling',
 ]
 python_version = "3.9"
 warn_unused_configs = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,9 @@ profile = "black"
 
 [tool.mypy]
 files = "src"
+exclude = [
+    'decaylanguage\.modeling\.*',
+]
 python_version = "3.9"
 warn_unused_configs = true
 warn_unused_ignores = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,11 +14,7 @@ write_to = "src/decaylanguage/_version.py"
 profile = "black"
 
 [tool.mypy]
-files = [
-    "src/decaylanguage/decay",
-    "src/decaylanguage/dec",
-    "src/decaylanguage/utils",
-]
+files = "src"
 exclude = "src/decaylanguage/(data|modeling)"
 python_version = "3.9"
 warn_unused_configs = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,14 +14,14 @@ write_to = "src/decaylanguage/_version.py"
 profile = "black"
 
 [tool.mypy]
-files = "src"
-exclude = [
-    '^decaylanguage/modeling',
+files = [
+    "src/decaylanguage/decay",
+    "src/decaylanguage/dec",
+    "src/decaylanguage/utils",
 ]
 python_version = "3.9"
 warn_unused_configs = true
 warn_unused_ignores = true
-check_untyped_defs = true
 strict = true
 
 [[tool.mypy.overrides]]
@@ -30,10 +30,8 @@ module = [
     "graphviz",
     "pandas",
     "plumbum",
-    "decaylanguage._version",
 ]
 ignore_missing_imports = true
-
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,11 +14,6 @@ write_to = "src/decaylanguage/_version.py"
 profile = "black"
 
 [tool.mypy]
-files = [
-    "src/decaylanguage/decay",
-    "src/decaylanguage/dec",
-    "src/decaylanguage/utils",
-]
 exclude = "src/decaylanguage/(data|modeling)"
 python_version = "3.9"
 warn_unused_configs = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,7 @@ files = [
     "src/decaylanguage/dec",
     "src/decaylanguage/utils",
 ]
+exclude = "src/decaylanguage/(data|modeling)"
 python_version = "3.9"
 warn_unused_configs = true
 warn_unused_ignores = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,9 +15,11 @@ profile = "black"
 
 [tool.mypy]
 files = "src"
-python_version = "3.6"
+python_version = "3.9"
 warn_unused_configs = true
+warn_unused_ignores = true
 check_untyped_defs = true
+strict = true
 
 [[tool.mypy.overrides]]
 module = [
--- a/src/decaylanguage/dec/dec.py
+++ b/src/decaylanguage/dec/dec.py
@@ -43,7 +43,18 @@ import re
 import warnings
 from io import StringIO
 from itertools import zip_longest
-from typing import Any, Iterable, Optional, Type, TypeVar, Union, no_type_check
+from typing import (
+    Any,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    no_type_check,
+)
 
 from lark import Lark, Tree, Visitor
 from particle import Particle
@@ -96,7 +107,7 @@ class DecFileParser:
         """
         self._grammar: Optional[str] = None  # Loaded Lark grammar definition file
         self._grammar_info: Optional[
-            dict[str, Any]
+            Dict[str, Any]
         ] = None  # Name of Lark grammar definition file
 
         # Name(s) of the input decay file(s)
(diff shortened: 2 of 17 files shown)


pyproject.toml
```diff
@@ -15,9 +15,14 @@ profile = "black"
 
 [tool.mypy]
 files = "src"
-python_version = "3.6"
+exclude = [
+    'decaylanguage\.modeling\.*',
```
eduardo-rodrigues, line 19: @henryiii, do you happen to know why the checking is being done on the modeling submodule as well when I'm trying here to ignore it? This line  hence isn't correct, somehow.

henryiii: This is not a valid regular expression (the star at then end is not preceded by anything to have 0 or more of). I think it should be

```suggestion
    '^decaylanguage/modeling',
```

It may need to be in an overrides too if it's being imported from files you are checking. If you are excluding the file from being an input file, but it's still being imported somewhere, it's still being checked, at least partially. You may want:

```toml
[[tool.mypy.overrides]]
module = "decaylanguage.modeling.*"
# Disable whatever it's struggling with here
```

Final]y, pre-commit might be passing these files by name if you haven't excluded them there.

eduardo-rodrigues: Let's see, thanks.

pyproject.toml
```diff
@@ -15,9 +15,14 @@ profile = "black"
...
+]
+python_version = "3.9"
 warn_unused_configs = true
+warn_unused_ignores = true
 check_untyped_defs = true
+strict = true
```
henryiii, line 25: I would not recommend starting with `strict=True` - that's great eventually, but it's probably better to go in stages. Here's the current list (in flag form, because that's what you get in `mypy --help`): `--warn-unused-configs, --disallow-any-generics, --disallow-subclassing-any,
                            --disallow-untyped-calls, --disallow-untyped-defs, --disallow-incomplete-defs, --check-untyped-defs, --disallow-untyped-
                            decorators, --no-implicit-optional, --warn-redundant-casts, --warn-unused-ignores, --warn-return-any, --no-implicit-
                            reexport, --strict-equality`.

You can also apply strict (or these finer-grained settings) to just one submodule at a time, I bet that's what you actually want to do.

But if it's already nearly passing strict, great! :)

src/decaylanguage/decay/decay.py
```diff
@@ -6,11 +6,21 @@
...
+    TypeVar,
+    Union,
+    no_type_check,
+)
 
 from ..utils import charge_conjugate_name
 
 
-class DaughtersDict(Counter):
+class DaughtersDict(Counter[str]):
```
henryiii, line 23: This is likely not subclass able until 3.8 or 3.9 - in fact, the ability to customize `__mro_entries__` to allow non-classes here, and the ability to have a `__class_getitem__`, wasn't added until 3.7. I'd do:

```python
if typing.TYPE_CHECKING:
    CounterStr = Counter[str]
else:
    CounterStr = Counter
```

Or similar.

src/decaylanguage/decay/decay.py
```diff
@@ -230,18 +256,23 @@ def from_dict(cls, decay_mode_dict):
         except Exception as e:
             raise RuntimeError("Input not in the expected format!") from e
 
-        return cls(bf=bf, daughters=daughters, **dm)
+        return cls(bf=bf, daughters=daughters, **dm)  # type: ignore
```
henryiii, line 259: I recommend including the error code, like `# type: ignore[error-code]`, as it explains why this is being ignored, and doesn't hide other errors too.

pyproject.toml
eduardo-rodrigues: Yeah, I've been working in batches but it's taken me time to read about  type checking and then implement it in the tens and tens of places. Not too far off a draft v0, actually.

src/decaylanguage/decay/decay.py
eduardo-rodrigues: Sounds good. Have not yet learned about this one :D.

henryiii: The codes are printed in brackets at the end of the line (in a different color):

```
src/decaylanguage/modeling/amplitudechain.py:46: error: Need type annotation for "err"  [var-annotated]
```

Not sure how to get it to ingore modeling yet. It's stubbornly checking it.

eduardo-rodrigues: Hmm. Tomorrow I will try and push forward with the remaining bits and will ping you ... Thanks already for starting to look 👍.

henryiii:
I've pushed a way to limit the checking a bit. It's not checking the base directory now (init, main, and version), but that's likely a lot more manageable. I couldn't get excluding to work, but opting-in does.

src/decaylanguage/decay/decay.py
eduardo-rodrigues: OK, will commit this in the next batch.

eduardo-rodrigues:
OK @henryiii, there are some errors left but I'm mostly done and would welcome your review and help with the remaining errors.

src/decaylanguage/decay/viewer.py
```diff
@@ -186,25 +186,25 @@ def has_subdecay(ds: list[Any]) -> bool:
         iterate_chain(sc)
 
     @property
-    def graph(self) -> graphviz.dot.Digraph:
+    def graph(self) -> graphviz.Digraph:
```
henryiii, line 189: There is no Digraph in `graphviz.dot`. It's in `graphvis.graphs`, but it's also exported to the top level, so let's just use that one. This was also incorrect in the README.

src/decaylanguage/dec/dec.py
```diff
@@ -1190,8 +1190,10 @@ def get_decays(parsed_file: Optional[str]) -> list[Tree]:
 
     try:
         return list(parsed_file.find_data("decay"))
-    except Exception:
-        RuntimeError("Input parsed file does not seem to have the expected structure.")
+    except Exception as err:
+        raise RuntimeError(
+            "Input parsed file does not seem to have the expected structure."
+        ) from err
```
henryiii, line 1196: MyPy was complaining because these functions were supposed to return something, but they were able to complete and return None since errors must be raised - simply creating an error object does nothing, and hides the original error.

I've also made flake8-bugbear happy about explicitly indicating that this error was caused by the other one. (That's the default in Python 3, but this verifies the intent)

eduardo-rodrigues: Oh, good eyes! Indeed I had not seen the `as err` missing in `except Exception as err` 👍 .

henryiii: The problem is `raise` was missing. :D The `as err` and `from err` was just a minor addition to make it clear that we are not hiding the original error that bugbear wanted. `from None` would hide the original error.

src/decaylanguage/dec/dec.py
```diff
@@ -96,7 +107,7 @@ def __init__(self, *filenames: Iterable[str]) -> None:
         """
         self._grammar: Optional[str] = None  # Loaded Lark grammar definition file
         self._grammar_info: Optional[
-            dict[str, Any]
+            Dict[str, Any]
```
henryiii, line 110: As long as we support Python 3.6 (which hit EOL last month, so probably not long?), we are stuck with using `Dict[str, str]` instead of `dict[str, str]` (along with `Union[str, int]` instead of `str | int`, etc.) :'(

src/decaylanguage/utils/utilities.py
```diff
@@ -45,12 +45,13 @@ def split(x):  # type: ignore [no-untyped-def]
     return out
 
 
-def filter_lines(matcher, inp):  # type: ignore [no-untyped-def]
+def filter_lines(
+    matcher: Pattern[str], inp: List[str]
+) -> Tuple[List[Dict[str, Union[str, Any]]], List[str]]:
```
henryiii, line 50: Adding type hints here, since they were missing. If you want to see the type hints for the standard library, like `re`, those are at https://github.com/python/typeshed/tree/master/stdlib - checking that is very helpful. That's where I found `Pattern`, and that it was in `typing`.

src/decaylanguage/dec/dec.py
eduardo-rodrigues: Thanks. Indeed I looked at most recent doc when learning and did not look at support of old versions ;-). Yes, I don't think 3.6 will need to stay for long. In fact I know of nobody still on 3.6.

src/decaylanguage/utils/utilities.py
```diff
@@ -45,12 +45,13 @@ def split(x):  # type: ignore [no-untyped-def]
...
     """
     Filter out lines into new variable if they match a regular expression
     """
-    output = [
-        matcher.match(ln).groupdict() for ln in inp if matcher.match(ln) is not None
-    ]
+    matches = (matcher.match(ln) for ln in inp)
+    output = [match.groupdict() for match in matches if match is not None]
```
henryiii, line 55: Reworked this to make mypy happy (and it's technically a line shorter now ;) ) - mypy doesn't know that `matcher.match(ln)` is the same when you run it twice. Without Python 3.8 walrus operator, the other way to fix it was to pull it out into a generator expression to avoid the double call, which I did.

eduardo-rodrigues: Nice.

henryiii: FYI, the walrus version would have been:

```python
output = [
    match.groupdict() for ln in inp if (match := matcher.match(ln)) is not None
]
```

As is often the case, I think the two line version is more readable than the walrus.

src/decaylanguage/dec/dec.py
eduardo-rodrigues: That will be nice indeed, and less to maintain.

src/decaylanguage/utils/utilities.py
eduardo-rodrigues: Indeed not too bad but the 2-line version might actually be quicker to read.

src/decaylanguage/utils/utilities.py
eduardo-rodrigues: Thank you for the pointer!

eduardo-rodrigues:
Thank you again for your help 👍 ! Feel freel to approve if happy (as agreed, there will be a follow-up but this is already a great improvement in the package, long due).

henryiii:
Pushed one more commit, pulling out some local functions and removing the function-wide type ignores. Please take a look at that commit, then we can merge. There should be a further followup, cleaning out a few more type ignores (or at least enforcing them to always have an error code).

henryiii (approved):

Issue #218, opened 2022-01-20
Test notebooks in the CI


eduardo-rodrigues:
Just noticed that this is not done here whereas we do it in `Particle`, which is optimal to keep docs up-to-date.

PR #219, opened 2022-01-20
Test notebooks in the CI

eduardo-rodrigues:
Closes https://github.com/scikit-hep/decaylanguage/issues/218.

A check of `AmpGen2GooFit D2K3p.ipynb` is skipped, see discussion below.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,9 @@ jobs:
     - name: Install notebook requirements
       run: python -m pip install nbconvert jupyter_client ipykernel
 
+    - name: Install dot
+      run: sudo apt-get install graphviz
+
     - name: Install pydot
       run: python -m pip install pydot
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,3 +42,23 @@ jobs:
     - name: Test coverage with Codecov
       if: "runner.os != 'Windows' && matrix.python-version != 3.7"
       uses: codecov/codecov-action@v2
+
+
+  notebooks:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - uses: actions/setup-python@v2
+
+    - name: Install package
+      run: python -m pip install -e .[test]
+
+    - name: Install notebook requirements
+      run: python -m pip install nbconvert jupyter_client ipykernel
+
+    - name: Run the notebooks inplace
+      run: jupyter nbconvert --execute --inplace notebooks/*.ipynb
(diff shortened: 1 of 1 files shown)


eduardo-rodrigues:
OK, so adding this check is immediately useful since a notebook is not up-to-date and needs fixing.

eduardo-rodrigues:
Hey @henryiii, I see a failure in your AmpGen related notebook `AmpGen2GooFit D2K3p.ipynb`. Can you have a look at what modification you might want? Seensa notebook cell should be either adapted or removed.

eduardo-rodrigues:
For reference the issue is that several parsed particles now have unknown spin, and the matching via name seems to have changed behaviour. Indeed in the past the code
```
for spintype, c in zip(SpinType, colors):
    ps = [
        c | format(str(p), "11")
        for p in GooFitChain.all_particles
        if p.spin_type == spintype
    ]
    print(c & colors.bold | "{:>12}:".format(spintype.name), *ps)
```
yielded
```
      Scalar: KPi20       PiPi30      PiPi00      PiPi10      KPi10       PiPi20      KPi00      
PseudoScalar: D0          pi-         pi+         K(1460)~-   K-         
      Vector: omega(782)0 K(892)~0    rho(770)0   rho(1450)0 
       Axial: K(1)(1270)~- a(1)(1260)+ K(1)(1400)~-
      Tensor: K(2)~-     
PseudoTensor:
     Unknown:
```
, as in the notebook presently in master, whereas now one gets
```
      Scalar: PiPi1       KPi10
PseudoScalar: pi+         D0          pi-         K-
      Vector: K*(892)~0   rho(1450)0  omega(782)  rho(770)0
       Axial: PiPi3       K(1)(1400)- K(1)(1270)- a(1)(1260)+
      Tensor: K(2)*(1430)-
PseudoTensor:
     Unknown: K(1460)-
  NonDefined: PiPi00      PiPi2       KPi00       KPi20
```
running the notebook.

Does this help you?

henryiii:
Ahh, yes, that helps. There's actually a test for exactly this, but it was disabled in #20. Particle originally had the extended PDG additions; we probably need to load them ourselves now.

eduardo-rodrigues:
> Ahh, yes, that helps. There's actually a test for exactly this, but it was disabled in #20. Particle originally had the extended PDG additions; we probably need to load them ourselves now.

Further on this. so the solution to your problem is not that trivial. For a start you create those spin factor "string names" that depend on SpinType. But that's something undefined for anything but mesons. And it turns out that the PDG IDs you attribute to some special particles defined in your file effectively correspond to a PDG ID for baryons, hence the `SpinType.NonDefined`. There are various ways to go around the problem but that's well beyond CI checks of notebooks ;-).

To be pragmatic I am now ignoring `AmpGen2GooFit D2K3p.ipynb` in the checks. That can be reverted in the future once the problem with the special particles is solved at source. I hope you agree with this approach.

This is therefore ready for review and approval.

henryiii:
Sounds good. We are already ignoring (since #20 IIRC) the matching test.

henryiii (approved):

henryiii:
Sounds good. We are already ignoring (since #20 IIRC) the matching test.

PR #233, opened 2022-04-07
Enhance DecFileParser.print_decay_modes function

eduardo-rodrigues:
I still need to add examples and tests ...

This enhancement may be of interest to you, @yipengsun.

### DIFF
--- a/src/decaylanguage/dec/dec.py
+++ b/src/decaylanguage/dec/dec.py
@@ -692,12 +692,12 @@ All but the first occurrence will be discarded/removed ...""".format(
     def print_decay_modes(
         self,
         mother: str,
-        pdg_name: Optional[bool] = False,
-        print_model: Optional[bool] = True,
-        display_photos_keyword: Optional[bool] = True,
-        ascending: Optional[bool] = False,
-        normalize: Optional[bool] = True,
-        normalization_bf: Optional[float] = 1,
+        pdg_name: bool = False,
+        print_model: bool = True,
+        display_photos_keyword: bool = True,
+        ascending: bool = False,
+        normalize: bool = True,
+        normalization_bf: float = 1.0,
     ) -> None:
         """
         Pretty print of the decay modes of a given particle.
@@ -729,7 +729,7 @@ All but the first occurrence will be discarded/removed ...""".format(
         """
         # Following does not catch the case "normalization_bf == 1 and not normalize"
         # but that's harmless anyway ;-)
-        if (normalization_bf != 1) and not normalize:
+        if (normalization_bf != 1.0) and not normalize:
             raise RuntimeError(
                 "Be consistent with normalize and normalization_bf arguments!"
             )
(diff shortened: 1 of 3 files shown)


src/decaylanguage/dec/dec.py
```diff
@@ -742,7 +756,16 @@ def print_decay_modes(
...
         ls = [(bf, ls_attrs_aligned[idx]) for idx, bf in enumerate(ls_dict)]
         ls.sort(key=operator.itemgetter(0), reverse=(not ascending))
-        norm = sum(bf for bf, _ in ls) if normalize else 1
+        norm = 1
+        if normalize:
+            # Get the largest branching fraction
+            i = -1 if ascending else 0
+            # Either normalize to 1 or normalization_bf
+            norm = (
+                ls[i][0] / normalization_bf
```
yipengsun, line 765: I'm a bit confused about this line. It seems that if a normalization constant is explicitly specified, the actual normalization would be `<largest_br_frac> / <specified_normalization>`.

What this means is that after normalization, the largest branching fraction will be `<specified_normalization>`? Am I interpreting this correctly?

Naively I'd just write `norm = normalization_bf if normalization_bf != 1 else sum(bf for bf, _ in ls)` but I have forgot the technical details of the code and might be completely wrong on this.

yipengsun: I re-read the doc string, and it seems like this is the intended behavior. I think it makes sense to do it like this.

src/decaylanguage/dec/dec.py
```diff
@@ -692,11 +692,12 @@ def _decay_mode_details(
...
-        pdg_name: Optional[bool] = False,
-        print_model: Optional[bool] = True,
-        display_photos_keyword: Optional[bool] = True,
-        ascending: Optional[bool] = False,
-        normalize: Optional[bool] = True,
+        pdg_name: bool = False,
+        print_model: bool = True,
+        display_photos_keyword: bool = True,
+        ascending: bool = False,
+        normalize: bool = False,
```
eduardo-rodrigues, line 699: FYI @yipengsun the normalisation is now set to False so that by default no values are "massaged". I think this is potentially clearer and more the expected behaviour,

src/decaylanguage/dec/dec.py
eduardo-rodrigues: As you can see above I added more doc and in particular a bunch of examples.

Yes the code is correct for what it is meant to do. But I agree that it can be a bit misleading because `normalization_bf` is effectively a scaling factor where as setting ` normalize=True` does normalise all numbers to a sum of unity. Those are different things. Maybe I should rather have two separate options `normalize` and `scale`, which cannot be used at the same time? I think I will go for that but am welcome to opinions. Thanks.

yipengsun: I think separate the `normalize` and `scale` as you described is more clear indeed.

eduardo-rodrigues: Yeah, seems even better now that it is implemented :-).

PR #234, opened 2022-04-09
DecFileParser - typing and function signature simplifications

eduardo-rodrigues:


eduardo-rodrigues:
Some obvious improvements in logic and simplifications/fixes concerning typing. Nothing major. I will merge Monday so that you have one day in case you want to have a look, @henryiii.

PR #263, opened 2022-10-07
feat: move to hatchling

henryiii:
See particle.


eduardo-rodrigues (approved):
Thank you Henry for this kind of updates in various packages.

PR #270, opened 2022-10-27
ModelAlias support and update of Belle II Decayfile

sognetic:
Hi,

picking up issue #100 I have implemented this in the grammar and added a Lark Transformer to add the correct models and options in the parse tree. I have also added the decay file used by Belle II (from the open source repository [here](https://github.com/belle2/basf2/blob/main/decfiles/dec/DECAY_BELLE2.DEC)) and added some decay models to the grammar to parse the current file.

The solution(s) I have found are not very elegant IMHO, especially adding white space/semicolons to the terminals is admittedly not great but I didn't see any other way to allow the model alias to contain the model name as a substring (as is the case in our decfile). As you can see I have added a second Lark Transformer (couldn't get it to work with a Visitor) to strip the white space and semicolons from the model name terminals.

All tests ran successfully after some slight changes needed to follow the new format of the Lark grammar. I'll happily add a test of the model alias functionality if you want.
Let me know what you think and thanks for this nice tool! We're definitely looking forward to using it in Belle II.

P.S.: What do you think about editing the Terminals needed for the model names dynamically with Lark's `edit_terminals` functionality? This way you could only maintain the enum for the model name and pass that dynamically to the parser. I'm not sure if this will work but if you like this idea I'll try it out and open a separate PR.


eduardo-rodrigues:
Hello @sognetic. First of all, thank you very much for picking this task up 👍 ! That's great to see and I know the functionality is something needed for Belle II.

I will look at this properly asap but it may not be this week. For now, really quickly: I see you add a new master .DEC file for Belle II. Is that replacing the existing one at https://github.com/scikit-hep/decaylanguage/blob/master/src/decaylanguage/data/DECAY_BELLE2.DEC? It looks as if you are adding a full new file, hence the question.

Back soon.

eduardo-rodrigues:
Yes, tests are really important :-).

sognetic:
> Hello @sognetic. First of all, thank you very much for picking this task up +1 ! That's great to see and I know the functionality is something needed for Belle II.

Yes, we (I) would really like to have the ability to parse our decfile in Belle II so I've been trying to find some time to implement this for quite some time.

> 
> I will look at this properly asap but it may not be this week. For now, really quickly: I see you add a new master .DEC file for Belle II. Is that replacing the existing one at https://github.com/scikit-hep/decaylanguage/blob/master/src/decaylanguage/data/DECAY_BELLE2.DEC? It looks as if you are adding a full new file, hence the question.

No, this decfile just replaces the existing one. I think the changes are just too large for the diff to work but I don't think it makes sense to deviate from the file we're actually using.

> Yes, tests are really important :-).

Okay, then I'll add one.

eduardo-rodrigues:
Thanks for the quick response. OK, good to know - clearly we want in the package the latest and greatest master decfile for Belle II!

sognetic:
Hi @eduardo-rodrigues , I have now added two tests to check the cleanup of the model names necessary after parsing, one test of the parsing of the ModelAlias definitions (I adapted one of the existing decfiles for this, this seemed fitting) and two tests (one on the level of decfiles, one directly with Lark Trees) for the Transformer which replaces the model aliases with the correct model names and options. 
Do you think this provides adequate test coverage of the changes or do you have anything else in mind?

eduardo-rodrigues:
I will finalise here this week. Just because I added all-contributors to the package, see the README, and will add you here, can you rebase to be on the HEAD? Thanks.

sognetic:
> I will finalise here this week. Just because I added all-contributors to the package, see the README, and will add you here, can you rebase to be on the HEAD? Thanks.

I've merged in the master and would prefer to rebase at the end of the review so I only have to force-push once (if that's fine with you).

eduardo-rodrigues:
Of course, no problem at all.

eduardo-rodrigues:
> P.S.: What do you think about editing the Terminals needed for the model names dynamically with Lark's edit_terminals functionality? This way you could only maintain the enum for the model name and pass that dynamically to the parser. I'm not sure if this will work but if you like this idea I'll try it out and open a separate PR.

To be honest I have not looked into Lark's edit_terminals so far and was not aware of its existence. Excellent if you can open a separate PR once this is merged and/or create a little issue to explain what you propose to do. Thanks a lot!

eduardo-rodrigues:
Quick question - have you timed the parsing of the LHCb and Belle II master .DEC files with all your enhancements in? It would be great to be reassured that parsing time stayed fine.

src/decaylanguage/dec/dec.py
```diff
@@ -196,11 +196,22 @@ def parse(self, include_ccdecays: bool = True) -> None:
         parser = Lark(
             self.grammar(), parser=opts["parser"], lexer=opts["lexer"], **extraopts
         )
-
         self._parsed_dec_file = parser.parse(self._dec_file)
 
+        # Strip whitespace and semicolons (necessary for LALR(1) grammar) from model names
+        self._parsed_dec_file = ModelNameCleanup().transform(self._parsed_dec_file)
```
eduardo-rodrigues, line 202: This is the first of 2 transformers. For small files there's no issue but for master .DEC files it may add some non-negligible time to navigate twice. Let's see as I've just asked you to time a couple of examples ... If timing increases non-negligibly, would it be possible to do these 2 transformers in one? Fine if this goes as a future endeavour - I'm also very keen on seeing this PR go in and released.

src/decaylanguage/dec/enums.py
```diff
@@ -95,4 +96,5 @@ class PhotosEnum(IntEnum):
     "VVP",
     "VVS_PWAVE",
     "YMSTOYNSPIPICLEO",
```
eduardo-rodrigues, line 98:
```suggestion
    "YMSTOYNSPIPICLEOBOOST",
    "YMSTOYNSPIPICLEO",
```

src/decaylanguage/dec/enums.py
```diff
@@ -95,4 +96,5 @@ class PhotosEnum(IntEnum):
...
+    "YMSTOYNSPIPICLEOBOOST",
```
eduardo-rodrigues, line 99:
```suggestion
```

eduardo-rodrigues: Unless it changed the longer string needs to go first otherwise the parser only finds the substring. This is easily checked with a mini test file using the longer name as model, and see if it is found.

src/decaylanguage/data/decfile.lark
```diff
@@ -49,12 +52,18 @@ model_options : (value | LABEL | _NEWLINE)+
...
+
+MODEL_NAME_AND_WS.2: MODEL_NAME _WS+
+MODEL_NAME_AND_SC.2: MODEL_NAME _WS* _SEMICOLON+
+
+MODEL_NAME : "BaryonPCR"|"BCL"|"BGL"|"BT02PI_CP_ISO"|"BTO3PI_CP"|"BTOSLLALI"|"BTOSLLBALL"|"BTOXSGAMMA"|"BTOXSLL"|"CB3PI-MPP"|"CB3PI-P00"|"D_DALITZ"|"ETAPRIME_DALITZ"|"ETA_DALITZ"|"ETA_FULLDALITZ"|"ETA_PI0DALITZ"|"FLATQ2"|"GENERIC_DALITZ"|"GOITY_ROBERTS"|"HELAMP"|"HQET3"|"HQET2"|"HQET"|"ISGW2"|"ISGW"|"LbAmpGen"|"LLSW"|"MELIKHOV"|"OMEGA_DALITZ"|"PARTWAVE"|"PHSP"|"PI0_DALITZ"|"PROPSLPOLE"|"PVV_CPLH"|"PYCONT"|"PYTHIA"|"SLBKPOLE"|"SLN"|"SLPOLE"|"SSD_CP"|"SSD_DirectCP"|"SSS_CP"|"SSS_CP_PNG"|"SSS_CPT"|"STS_CP"|"STS"|"SVP_CP"|"SVP_HELAMP"|"SVP"|"SVS_CP_ISO"|"SVS_CPLH"|"SVS_CP"|"SVS_NONCPEIGEN"|"SVS"|"SVV_CPLH"|"SVV_CP"|"SVV_HELAMP"|"SVV_NONCPEIGEN"|"SVVHELCPMIX"|"TAUHADNU"|"TAULNUNU"|"TAUSCALARNU"|"TAUVECTORNU"|"TSS"|"TVP"|"TVS_PWAVE"|"VLL"|"VSP_PWAVE"|"VSS_BMIX"|"VSS_MIX"|"VSS"|"VUB"|"VVPIPI"|"VVP"|"VVS_PWAVE"|"YMSTOYNSPIPICLEO"|"YMSTOYNSPIPICLEOBOOST"
```
eduardo-rodrigues, line 66:
```suggestion
MODEL_NAME : "BaryonPCR"|"BCL"|"BGL"|"BT02PI_CP_ISO"|"BTO3PI_CP"|"BTOSLLALI"|"BTOSLLBALL"|"BTOXSGAMMA"|"BTOXSLL"|"CB3PI-MPP"|"CB3PI-P00"|"D_DALITZ"|"ETAPRIME_DALITZ"|"ETA_DALITZ"|"ETA_FULLDALITZ"|"ETA_PI0DALITZ"|"FLATQ2"|"GENERIC_DALITZ"|"GOITY_ROBERTS"|"HELAMP"|"HQET3"|"HQET2"|"HQET"|"ISGW2"|"ISGW"|"LbAmpGen"|"LLSW"|"MELIKHOV"|"OMEGA_DALITZ"|"PARTWAVE"|"PHSP"|"PI0_DALITZ"|"PROPSLPOLE"|"PVV_CPLH"|"PYCONT"|"PYTHIA"|"SLBKPOLE"|"SLN"|"SLPOLE"|"SSD_CP"|"SSD_DirectCP"|"SSS_CP"|"SSS_CP_PNG"|"SSS_CPT"|"STS_CP"|"STS"|"SVP_CP"|"SVP_HELAMP"|"SVP"|"SVS_CP_ISO"|"SVS_CPLH"|"SVS_CP"|"SVS_NONCPEIGEN"|"SVS"|"SVV_CPLH"|"SVV_CP"|"SVV_HELAMP"|"SVV_NONCPEIGEN"|"SVVHELCPMIX"|"TAUHADNU"|"TAULNUNU"|"TAUSCALARNU"|"TAUVECTORNU"|"TSS"|"TVP"|"TVS_PWAVE"|"VLL"|"VSP_PWAVE"|"VSS_BMIX"|"VSS_MIX"|"VSS"|"VUB"|"VVPIPI"|"VVP"|"VVS_PWAVE"|"YMSTOYNSPIPICLEOBOOST"|"YMSTOYNSPIPICLEO"
```

tests/data/defs-aliases-chargeconj.dec
```diff
@@ -276,3 +276,13 @@ SetLineshapePW D_10 D*0 pi0 2
 SetLineshapePW D_10 D*+ pi- 2
 SetLineshapePW anti-D_10 anti-D*0 pi0 2
 SetLineshapePW anti-D_10 D*- pi+ 2
+
+
+# Model aliases used in Belle II
+ModelAlias SLBKPOLE_DtoKlnu SLBKPOLE 1.0 0.303 1.0 2.112;
```
eduardo-rodrigues, line 282: Would a line such as `ModelAlias SLBKPOLE_DtoKlnu SLBKPOLE 1.0 0.303 1.0 2.112 ;` work? Then add a test?

tests/data/test_DtoKlnu.dec
```diff
@@ -0,0 +1,13 @@
+# Example decay for testing purposes:
+# This checks whether ModelAliases are parsed and replaced correctly during parsing
```
eduardo-rodrigues, line 2:
```suggestion
# This checks whether model aliases are parsed and replaced correctly during parsing
```

tests/data/test_DtoKlnu.dec
```diff
@@ -0,0 +1,13 @@
...
+# and if this is also works with variable definitions.
```
eduardo-rodrigues, line 3:
```suggestion
# and if this also works with variable definitions.
```

tests/dec/test_dec.py
```diff
@@ -191,6 +193,25 @@ def test_missing_global_photos_flag():
     assert not p.global_photos_flag()
 
 
+def test_model_aliases():
```
eduardo-rodrigues, line 196: Totally cosmetics but can you move this test to next to the related test `test_aliases_parsing()` and while at it follow the same name "convention?

tests/dec/test_dec.py
```diff
@@ -191,6 +193,25 @@ def test_missing_global_photos_flag():
...
+    p = DecFileParser(DIR / "../data/defs-aliases-chargeconj.dec")
+    p.parse()
+
+    assert len(p.dict_model_aliases()) == 7
+    assert p.dict_model_aliases()["SLBKPOLE_DtoKlnu"] == [
```
eduardo-rodrigues, line 201: I'm confused here. I understood that `dict_model_aliases()` is providing a dict, not a Lark object "unmassaged". I was then expecting something similar to what you see for example in
```
def test_pythia_definitions_parsing():
    p = DecFileParser(DIR / "../data/defs-aliases-chargeconj.dec")
    p.parse()

    assert p.dict_pythia_definitions() == {
        "ParticleDecays:mixB": "off",
        "Init:showChangedSettings": "off",
        "Init:showChangedParticleData": "off",
        "Next:numberShowEvent": 0.0,
    }


def test_jetset_definitions_parsing():
    p = DecFileParser(DIR / "../data/defs-aliases-chargeconj.dec")
    p.parse()

    assert p.dict_jetset_definitions() == {
        "MSTU": {1: 0, 2: 0},
        "PARU": {11: 0.001},
        "MSTJ": {26: 0},
        "PARJ": {21: 0.36},
    }
```

tests/dec/test_dec.py
```diff
@@ -288,6 +309,34 @@ def test_decay_model_parsing_with_variable_defs():
...
+    """
+    In this example the decay model details are "SLBKPOLE_DtoKlnu",
+    where SLBKPOLE_DtoKlnu stands for an alias whose value is defined via the statement
+    "ModelAlias SLBKPOLE_DtoKlnu SLBKPOLE 1.0 param1;" (semicolon matters here).
+    "param1" is here defined via the statement "Define param1 -0.303".
+    The parser should recognise this and return SLBKPOLE as the model name and [1.0, -0.303] as the model parameters.
+    """
+    p = DecFileParser(DIR / "../data/test_DtoKlnu.dec")
+    p.parse()
+    assert p.dict_model_aliases() == {
```
eduardo-rodrigues, line 322: Ditto.

tests/dec/test_dec.py
```diff
@@ -621,11 +785,12 @@ def test_lark_file_model_list_consistency():
     with open(filename) as lark_file:
         lines = lark_file.readlines()
         for line in lines:
-            if "MODEL_NAME.2" in line:
+            if "MODEL_NAME :" in line:
                 break
         models = line.split(":")[1].strip(" ").strip("\n").split('"|"')
         models = [m.strip('"') for m in models]
 
+        assert set(models) == set(known_decay_models)
```
eduardo-rodrigues, line 793: Why isn't the next line enough?

eduardo-rodrigues:
This last batch of commits with tests look very nice. I left some comments but it's all rather minor in the end. I'm going to approve and run in case the CI catches something my eyes did not. We're basically there 👍!

src/decaylanguage/dec/enums.py
sognetic: I don't think this is necessary anymore. The grammar requires now that the model name is followed by at least one white space character or a semicolon (I think that's more or less what evtgen does but don't quote me on that) so it will match the correct model name regardless. However I will run some tests to ensure that what I'm saying is true.

src/decaylanguage/data/decfile.lark
sognetic: As above: I don't think that's necessary anymore.

tests/dec/test_dec.py
sognetic: I thought it might be useful in interpreting the test output to know if a model has simply been forgotten or if the order is wrong.

src/decaylanguage/dec/enums.py
eduardo-rodrigues: Cool if that's the case now. That will be an improvement in spite of the stricter requiremen of a WS or SC. Better check but if OK then simply reject my changes. Thanks.

src/decaylanguage/data/decfile.lark
eduardo-rodrigues: Ditto for my reply then :-).

tests/dec/test_dec.py
eduardo-rodrigues: Oh, but if forgotten the list would not match. Same for the order. So that's already covered unless I'm really missing a use-case. If yes then can you give me an example with [1,2,3,4]?

It is not a big issue to have the extra line but nicer not to have redundant stuff.

src/decaylanguage/dec/dec.py
sognetic: I've tested this now and there is some significant slowdown:

LHCb: 18% slower
Belle II: 43% slower (using the old decay file)

For us being able to use this tool at all is great so I wouldn't see this as unacceptable but I see that the LHCb perspective might be somewhat different here.

Combining the two transformers is not straightforward as the first transformer also replaces the semicolon / whitespace in the second one. I'm sure there are ways to decrease the performance penalty but that might require some time on my end.

tests/dec/test_dec.py
sognetic: You're right that both cases would be covered with the last line. I added this line as the test was failing during debugging and I wasn't sure if I had only messed up the order. I agree that it's redundant and have removed it.

src/decaylanguage/dec/dec.py
eduardo-rodrigues: Thanks a lot for the speedy reaction.

OK, it does have a clear impact but not the end of the world either. We can probably work on improving in the future ...

I guess the larger difference for the Belle II file is simply stating that model aliases add a clear time penalty in parsing.

No worries, this is not unacceptable at all. I should say that I started on this really for anyone using dec file, not specifically for LHCb. In that sense, to me it is *extremely important and relevant* to have Belle II colleagues on board. If you are happy then I'm happy that this tool can serve us all :-).

Totally understand that the merging is not straightforward. I realised that after making my comment. No worries.

Good that we have these numbers in mind for any future reference.

src/decaylanguage/dec/enums.py
sognetic: This works, I've added a test to ensure this too.

tests/data/defs-aliases-chargeconj.dec
sognetic: Yes, this works. I've added a sub-test in the relevant function.

src/decaylanguage/dec/enums.py
eduardo-rodrigues: Many thanks.

tests/dec/test_dec.py
eduardo-rodrigues: I totally understand the good reason of avoiding a packing-unpacking sequence. But then one would separate what are internal manipulations and what are functions that provide info to the user in a friendly way. Maybe here you would need mostly internal functions and the public one would give a dict of alias:model names only. Let me know what you think.

Otherwise it seems we're good to go.

sognetic: That seems like a good idea, I'll split this into public and private methods.

sognetic: Okay, I've now done that. For some reason the ` _parsed_dec_file` in `DecFileParser` was annotated with `str` instead of `Tree` which mypy didn't accept. I've changed it to `Tree | None`, please let me know if there was a reason for the previous behavior and I'll revert this.

And thank you for finding time to review this!

eduardo-rodrigues: Nope, strange that mypy did not complain before. I won't investigate, though.

eduardo-rodrigues (approved):
I think this PR is ready 👍!

Thank you again for this great and highly awaited enhancement!

Let's follow-up anything with future contributions, as said in the discussions (there was at least a suggestion of further work).

eduardo-rodrigues:
@all-contributors please add @sognetic for code

eduardo-rodrigues:
@all-contributors please add @sognetic for code

eduardo-rodrigues:
Hello @sognetic, have you run all tests locally? It's rather odd the "failures" observed because of a deprecation warning that was not there in the previous CI tests run 2 days ago that use all the same dependencies as far as I can see. Not sure what to do with this. I'm tempted to merge if you can show the results of the tests locally.

eduardo-rodrigues:
Hey, line https://github.com/scikit-hep/decaylanguage/blob/master/.github/workflows/ci.yml#L48 does not seem right. Can you replace by `python-version: ${{ matrix.python-version }}` so that I do not go and fork&clone your stuff? Easier. We then see if that fixes at least the tests on everything but Python 3.11. Advance thanks.

sognetic:
> Hello @sognetic, have you run all tests locally? It's rather odd the "failures" observed because of a deprecation warning that was not there in the previous CI tests run 2 days ago that use all the same dependencies as far as I can see. Not sure what to do with this. I'm tempted to merge if you can show the results of the tests locally.

Hi @eduardo-rodrigues, I have run all tests (just running `nox` right), but only with python3.8 as that is all I have easily available. I can try to find a machine with more versions of the interpreter if necessary.


> Hey, line https://github.com/scikit-hep/decaylanguage/blob/master/.github/workflows/ci.yml#L48 does not seem right. Can you replace by `python-version: ${{ matrix.python-version }}` so that I do not go and fork&clone your stuff? Easier. We then see if that fixes at least the tests on everything but Python 3.11. Advance thanks.

Okay, done

eduardo-rodrigues:
Great,I like it much better when things are green!

I see the failure for Python 3.11 has been fixed in Lark for versions 1.x whereas we are on 0.12, the latest to support Python 2. This is all details independent from this PR, so let's bypass the failure.

Keep in touch and do not forget the little follow-ups. Advance thanks.

sognetic:
> Great,I like it much better when things are green!
> 
Yes, I totally agree. Green is better and makes me sleep much better.
> 
> Keep in touch and do not forget the little follow-ups. Advance thanks.

Thank you for the great review and I'll try to find some time for these follow-ups! I like this package and hope we can build some nice things with it in Belle II.

sognetic:
Hi @eduardo-rodrigues , this is great to hear! I'm advertising the tool already as much as possible, I think we can build some very nice validation things with it. I have already dropped a star of course but will encourage everybody else to do so as well.

Issue #280, opened 2022-11-08
Adapt to Lark 1.x


eduardo-rodrigues:
Indeed we have been depending on lark-parser 0.x series and the new lark 1.x releases are Python 3 only.

Pip installs become `pip install lark` instead of `pip install lark-parser`, see https://github.com/lark-parser/lark.

PR #282, opened 2022-11-08
Adapt to Python 3 only Lark 1.x

eduardo-rodrigues:
Closes https://github.com/scikit-hep/decaylanguage/issues/280.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Required and compatibility dependencies will be automatically installed by pip.
 -   [pandas](https://pandas.pydata.org/): Tabular data in Python
 -   [attrs](https://github.com/python-attrs/attrs): DataClasses for Python
 -   [plumbum](https://github.com/tomerfiliba/plumbum): Command line tools
--   [lark-parser](https://github.com/lark-parser/lark): A modern parsing library for Python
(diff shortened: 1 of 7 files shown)


eduardo-rodrigues:
This is a no-brainer and fixes Python 3.11 without a need for a hack. Clean.

Any comments @henryiii? Else approve just so I know you're happy as well :-).

henryiii:
> fixes Python 3.11 without a need for a hack


Is something broken on Python 3.11?

eduardo-rodrigues:
> > fixes Python 3.11 without a need for a hack
> 
> Is something broken on Python 3.11?

The CI tests were failing on 3.11 with recent PRs because of a deprecation warning, see https://github.com/scikit-hep/decaylanguage/actions/runs/3393703868/jobs/5641353853. A fix would require a minor release in Lark for the o.x series. I then realised that the fix got implemented in the 1.x series and decided to go "2-in-1" simply trying to move to Lark 1.x :-).

henryiii:
Ahh, okay, yes, I remember that warning.

eduardo-rodrigues:
I will try and release DecayLanguage asap since Particle and DecayLanguage both have great updates in, for Belle II colleagues and cosmics!

PR #351, opened 2023-07-21
Simplify the decay file parsing grammar and implement more get methods for EvtGen keywords + unit tests

eduardo-rodrigues:
Hi @henryiii, I realised while working on get methods for recently introduced EvtGen  keywords that (1) some rules with more than one possible label cannot presently be disentangled (e.g. "ChangeMassMin" and "ChangeMassMax") and (2) the grammar can anyway be simplified, potentially making things slightly faster.

I will be adding more here but would welcome your opinion on these simplifications in case I am missing something - I think not and this has no cons, just pros. Advance thanks.

### DIFF
--- a/src/decaylanguage/data/decfile.lark
+++ b/src/decaylanguage/data/decfile.lark
@@ -18,7 +18,7 @@ setlsbw : "BlattWeisskopf" label value // Set Blatt-Weisskopf barrier factor for
 
 setlspw : "SetLineshapePW" LABEL LABEL LABEL INT // Redefine Partial Wave for label -> label label
 
-cdecay : "CDecay" label
+cdecay : "CDecay" LABEL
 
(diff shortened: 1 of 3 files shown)


eduardo-rodrigues:
FYI @admorris, this is a follow-up atop your recent work. I simplified the grammar overall and also had to make some improvements to be able to disentangle post-parsing between values for "ChangeMassMin" and "ChangeMassMax", for example. Tests have been added.

There's more to come (for example there are no get methods for now for "IncludeBirthFactor" or "IncludeDecayFactor"), but we're getting quite close to being done here.

I will leave this open until Monday or Tuesday in case you manage to have a look and have comments. Otherwise a thumbs-up will do if happy :-).

I hope this proves useful for your LHCb work.

Advance thanks.

PR #356, opened 2023-07-25
Enhance Lark parsing grammar to parse all LHCb dec files

eduardo-rodrigues:


eduardo-rodrigues:
FYI @admorris - to the best of my knowledge DecayLanguage now parses OK all 8000-ish LHCb decay files with this enhancement :-).

Issue #357, opened 2023-07-25
Enhance test suite to verify all EvtGen models


eduardo-rodrigues:
Ideally, all EvtGen models known to the package should be tested comprehensively. Not only the parsing succeeds but dedicated tests would ensure that all model parameters, including model aliases and model options, are all parsed correctly.

The presently known list of models is the following (items to be ticked off as dedicated model parsing correctness tests are added):

- [x] BaryonPCR
- [x] BC_SMN
- [x] BC_TMN
- [x] BC_VHAD
- [x] BC_VMN
- [x] BCL
- [x] BGL
- [x] BLLNUL
- [x] BNOCB0TO4PICP
- [x] BNOCBPTO3HPI0
- [x] BNOCBPTOKSHHH
- [x] BS_MUMUKK
- [x] BSTOGLLISRFSR
- [x] BSTOGLLMNT
- [ ] BT02PI_CP_ISO (no dec file available for testing from LHCb or Belle-II)
- [x] BTO3PI_CP
- [x] BTODDALITZCPK
- [x] BToDiBaryonlnupQCD
- [x] BTOSLLALI
- [x] BTOSLLBALL
- [x] BTOSLLMS
- [x] BTOSLLMSEXT
- [x] BTOVLNUBALL
- [x] BTOXSGAMMA
- [ ] BTOXELNU
- [x] BTOXSLL
- [x] BQTOLLLLHYPERCP
- [x] BQTOLLLL
- [x] CB3PI-MPP
- [x] CB3PI-P00
- [x] D_DALITZ
- [x] D_hhhh
- [x] D0GAMMADALITZ
- [x] D0MIXDALITZ
- [x] DToKpienu
- [x] ETAPRIME_DALITZ
- [x] ETA_DALITZ
- [x] ETA_FULLDALITZ
- [x] ETA_LLPIPI
- [x] ETA_PI0DALITZ
- [x] FLATQ2
- [x] FLATSQDALITZ
- [x] FOURBODYPHSP
- [ ] GENERIC_DALITZ (no dec file available for testing from LHCb or Belle-II)
- [x] GOITY_ROBERTS
- [x] HELAMP
- [x] HQET3
- [x] HQET2
- [x] HQET
- [x] ISGW2
- [x] ISGW
- [x] KS_PI0MUMU
- [x] Lb2Baryonlnu
- [x] Lb2plnuLCSR
- [x] Lb2plnuLQCD
- [x] LbAmpGen
[...]

PR #385, opened 2023-11-06
Refactored grammar to remove need for second transformer

sognetic:
Hi,

as promised (with some delay) I've had another think about the feature introduced in #100 which slowed down processing quite a bit by introducing an unnecessary transformer. I've removed this transformer again and refactored the grammar such that the `MODEL_NAME` terminal doesn't need to include the white space / semicolon anymore but just matches on a word boundary (`\b`). This simplifies the grammar quite a bit and speeds up parsing of the LHCb / Belle II decay files by 15%.
I've removed the tests corresponding to the removed transformer, I'm not sure if / how I should replace them with something in this PR. All other tests pass.

Let me know what you think!

EDIT: I don't know why the `pre_commit` check in the CI doesn't work, everything is okay locally.


eduardo-rodrigues:
Hello @sognetic, nice to hear back from you and thank you very much already for this new contribution 👍!

This PR pushed me to continue on task https://github.com/scikit-hep/decaylanguage/issues/357, see the latest https://github.com/scikit-hep/decaylanguage/pull/387 from today. One model in particular, rather unique in the way it span many lines, needed to be tested.

I will get to your PR early next week at the very latest. In the meantime can you rebase to (1) get some more important tests run and (2) get a fix for the pre-commit hook CI (failing for you as well abvove)? Thanks.

sognetic:
Hi @eduardo-rodrigues!
Wow, #357 seems like a huge effort but also extremely useful to corral the EvtGen model zoo. I've merged in the current master branch so hopefully that fixes the failing CI thing. Everything still passes locally.

I've also worked on a second feature which removes the need to maintain two lists of models in both `dec/enums.py` and the grammar definition itself. I'll open a PR with this once this one is merged since I've based it on the grammar changes already included here.

eduardo-rodrigues:
Sounds great!

Yeah, task https://github.com/scikit-hep/decaylanguage/issues/357 is a bit of a pain and repetitive work, but worth to ensure the parsing not only succeedes but also gives correctly parsed info. So far I've found 2-3 issues and made several improvements along the way. It does take time.

src/decaylanguage/data/decfile.lark
```diff
@@ -55,17 +55,13 @@ model_label : LABEL
 
 model_alias : "ModelAlias" model_label model
 
-model : (model_label _SEMICOLON | MODEL_NAME_AND_SC | MODEL_NAME_AND_WS model_options _SEMICOLON+ | MODEL_NAME_MULTILINE model_options _SEMICOLON+)
+model : (model_label  | MODEL_NAME model_options?) _SEMICOLON+
```
eduardo-rodrigues, line 58: Excellent job! I'm impressed with the simplification achieved here, especially that I remember playing quite some time with the grammar to be able to support everything as before and parse the multi-line model PTO3P!

eduardo-rodrigues: Out of curiosity, you simply used timeit to check the improvement in parsing speed? 15% is already very nice.

eduardo-rodrigues: To be checked but I think you can now also remove the line `%import common.NEWLINE` below.

eduardo-rodrigues (changes requested):
Brilliant 👍!

I'm requesting trivial changes but that's trivial.

src/decaylanguage/data/decfile.lark
sognetic: > Excellent job! I'm impressed with the simplification achieved here, especially that I remember playing quite some time with the grammar to be able to support everything as before and parse the multi-line model PTO3P!

Thank you! I think the issue was mostly caused by me not understanding Python regex syntax well enough when changing the Lark grammar last year so the simplification was only really possible by partially reverting this change.

> Out of curiosity, you simply used timeit to check the improvement in parsing speed? 15% is already very nice.

Exactly, I just ran timeit with 50 repetitions on the LHCb / Belle II main decfiles. The results were pretty similar for both files (not surprising I guess since the relevant `Transformer` just visited each model terminal).

> To be checked but I think you can now also remove the line `%import common.NEWLINE` below.

You're right, I've removed it and all tests still ran through without problems.

eduardo-rodrigues (approved):
Thanks again for this!

PR #390, opened 2023-11-09
Dynamically supply list of decay models to grammar

sognetic:
Hi @eduardo-rodrigues,

here's my second PR which replaces the hard-coded list of model names in the grammar with a placeholder which is filled dynamically with the list of models. I've marked the PR as a draft since there are a few points where I'm not sure what the way to handle things is. I'll comment on the relevant lines / changes in the PR and would be grateful for some input there.

All tests pass locally.


src/decaylanguage/dec/dec.py
```diff
@@ -195,7 +196,11 @@ def parse(self, include_ccdecays: bool = True) -> None:
 
         # Instantiate the Lark parser according to chosen settings
         parser = Lark(
-            self.grammar(), parser=opts["parser"], lexer=opts["lexer"], **extraopts
+            self.grammar(),
+            parser=opts["parser"],
+            lexer=opts["lexer"],
+            edit_terminals=_edit_model_name_terminals,
```
sognetic, line 202: A disadvantage of hard-coding it like this is that `edit_terminals` cannot be supplied via `extraopts` anymore. If you see a need to still be able to do that I'm sure I could try to come up with something that combines all callbacks somehow.

src/decaylanguage/dec/dec.py
```diff
@@ -1813,3 +1818,13 @@ def _str_to_bool(arg: str) -> bool:
...
         f"String {arg!r} cannot be converted to boolean! Only 'yes/no' accepted."
     )
+
+
+def _edit_model_name_terminals(t: TerminalDef) -> None:
+    """
+    Edits the terminals of the grammar to replace the model name placeholder with the actual names of the models.
+    """
+
+    modelstr = rf"(?:{'|'.join(re.escape(kdm) for kdm in sorted(known_decay_models, key=len, reverse=True))})"
```
sognetic, line 1828: I'm sorting the models since that's what Lark does. It also works without sorting but then a bunch of tests fail which seem to rely on this sorting.

tests/dec/test_known_decay_models.py
```diff
@@ -15,27 +15,6 @@
 DIR = Path(__file__).parent.resolve()
 
 
-def test_lark_file_model_list_consistency():
```
sognetic, line 18: I've removed this test since it doesn't really make sense anymore. I'm not sure how to replace it with something new though, any idea?

eduardo-rodrigues:
Hey. A was not aware of this new(?) feature from Lark. What are the constraints to use it in terms of dependencies and Python version support?

TBH I quite like the idea that one can trivially see from the definition file what models are available, and extend if necessary, see https://github.com/scikit-hep/decaylanguage/blob/master/README.md#advanced-usage. What would be the new way to achieve
```
dfp = DecFileParser('my_decfile.dec')
dfp.load_grammar('path/to/my_updated_decfile.lark')
dfp.parse()
...
```
discussed there with your changes? Apologies that I'm really loaded and did not look carefully, and I could probably answer myself. I will do soon unless you make it easier for me :-).

src/decaylanguage/dec/dec.py
eduardo-rodrigues: Hey. You mean if somebody was to fully define the parser? I added all that flexibility at the beginning but might remove it at some point. The fact is that I believe nobody has ever used it and there are no tests to ensure other Lark parsers do the job right, and wisdom says that if something is not tested, it is not in practice supported.

Is this what you meant?

tests/dec/test_known_decay_models.py
eduardo-rodrigues: The removal is logical given your move since there is no consistency to be checked between 2 "sources".

src/decaylanguage/dec/dec.py
sognetic: Yes, exactly. I mean if someone is currently passing `edit_terminals` to `load_grammar()` that would end up in `extraopts` and clash with this additional argument. I'm not sure anyone really does this but since `load_grammar()` is a public method I thought I'd check with you.

sognetic:
> Hey. A was not aware of this new(?) feature from Lark. What are the constraints to use it in terms of dependencies and Python version support?

As far as I can tell, this (imo not very well-documented) feature has existed since version `0.7.7` of Lark, being added after discussion in [this](https://github.com/lark-parser/lark/issues/406)  issue.
I'm not sure what this means for Python version support but would argue that a feature that was added four years ago should be relatively safe. The tests with all different versions pass after all.

> 
> TBH I quite like the idea that one can trivially see from the definition file what models are available, [...]

I agree that having a full description in the grammar is nice but would argue that the simplicity of only having to maintain one list beats that. Having a hard-coded list of strings in the grammar also feels kind of weird to me, a change of the grammar should indicate a fundamental change of the language, not just an update of the list of allowed models. While the weird decfile language makes this list kind of necessary (I really tried to come up with a way around hard-coding the models but couldn't come up with one) it's not really desirable in my opinion.

> [...] and extend if necessary, see https://github.com/scikit-hep/decaylanguage/blob/master/README.md#advanced-usage. What would be the new way to achieve
> 
> ```
> dfp = DecFileParser('my_decfile.dec')
> dfp.load_grammar('path/to/my_updated_decfile.lark')
> dfp.parse()
> ...
> ```
> 
> discussed there with your changes? 

If `path/to/my_updated_decfile.lark` doesn't contain `MODEL_NAME_PLACEHOLDER` I think the example should still work and would just use the hard-coded list of models in the grammar, right? 
We could add an optional argument to `load_grammar()` which could be used to supply an alternative list of models and the appropriate placeholder string.

> Apologies that I'm really loaded and did not look carefully, and I could probably answer myself. I will do soon unless you make it easier for me :-).

No worries, it's not an urgent feature really but I thought I'd just put it in a draft so I don't forget this exists and to start some discussion on whether this is something you'd want in `decaylanguage`.

src/decaylanguage/dec/dec.py
eduardo-rodrigues: Hmm. If I remember correctly, the main/real reason why I added `load_grammar` was to be able to easily define an updated grammar, typically for a new model, and then continue transparently. Now, if the models aren't in the grammar anymore, than makes far less sense.

It would be useful to see a couple of use-cases spelled out, see my comment https://github.com/scikit-hep/decaylanguage/pull/390#issuecomment-1804349189 below. The way you suggest, which is sounding good to me, we should then deprecate `load_grammar` with a reference to this PR so that people can understand the rationale. I don't think we need to give much advance warning - basically the next minor version 0.18.0 - since, as said already, I do not think anyone is in fact using this feature. Besides, we now are in a much much better situation of supporting all LHCb and Belle II models we are aware of as of today. It is unlikely that anyone using a fresh new model won't be able to wait 1-2 days max for a new release with the new model tested ;-).

src/decaylanguage/dec/dec.py
eduardo-rodrigues: There is some griddy pattern matching, I recall, which explained the sorting. For example, the parking might find XXX in model XXXYYY even if a model XXX also exists, and then the latter would not be parsed correctly. I may be mistaken but ...

eduardo-rodrigues:
Hello @sognetic, sorry for being a bit silent here. You may have seen that I have been very busy finalising the tests of the EvtGen models. I'm now basically done and you can see from https://github.com/scikit-hep/decaylanguage/pull/396 that some enhancements and fixes were very important to get in. I'm afraid you will need to rebase one last time, but for a good cause.

src/decaylanguage/dec/dec.py
eduardo-rodrigues: Hi. Let's try and get back to this.

Again, I understand why you are suggesting this PR. The removal of duplication sounds fair. Going in this direction, you need to remove the `load_grammar` method. This package is not used by tons of people, hence acceptable, I reckon.

Now, if we wanted to still give the flexibility for a user to provide a list of models, that would need to be done via a public method. What do you think?

eduardo-rodrigues:
I would suggest that we go ahead here with the removal and we then clean-up a bit. I see also that the notebook example needs an update ... Let me know what's your timeline, so that I adapt - I have a new PR to make to provide full coverage of Belle II models, so if you tell me you cannot work in the next few days I can make that MR and you then rebase on the full set of my PRs, else I will wait. Just let me know.

Thanks.

sognetic:
Okay, I wasn't sure whether your proposed enhancements had already finished or not. Unfortunately I cannot work on this in the next seven days but will pick it up afterwards. I'll gladly merge in any changes improving Belle II coverage, thank you for doing this!

src/decaylanguage/data/decfile.lark
```diff
@@ -58,10 +58,12 @@ model_alias : "ModelAlias" model_label model
...
+// model : model_generic
```
eduardo-rodrigues, line 65:
```suggestion
```

src/decaylanguage/data/decfile.lark
```diff
@@ -58,10 +58,12 @@ model_alias : "ModelAlias" model_label model
...
+// model_helamp : "HELAMP" (SIGNED_NUMBER SIGNED_NUMBER)+
```
eduardo-rodrigues, line 66:
```suggestion
```

src/decaylanguage/dec/dec.py
```diff
@@ -135,6 +137,9 @@ def __init__(self, *filenames: str | os.PathLike[str]) -> None:
         self._parsed_decays: None | (
             Any
         ) = None  # Particle decays found in the decay file
+        self._additional_decay_models: None | Iterable[
```
eduardo-rodrigues, line 140:
```suggestion

        self._additional_decay_models: None | Iterable[
```

src/decaylanguage/dec/dec.py
```diff
@@ -163,9 +168,8 @@ def parse(self, include_ccdecays: bool = True) -> None:
         Parse the given .dec decay file(s) according to the default Lark parser
         and specified options.
 
-        See method 'load_grammar' for how to explicitly define the grammar
-        and set the Lark parsing options. This method needs to be called
-        before 'parse' to override the parser and its options.
+        Use the method `load_additional_decay_models` before `parse` to load decay models
+        that might not yet be available in 'decaylanguage'.
```
eduardo-rodrigues, line 172:
```suggestion
        that might not yet be available in DecayLanguage.
```

src/decaylanguage/dec/dec.py
```diff
@@ -281,18 +294,80 @@ def load_grammar(
...
+        )
+
         if filename is None:
             filename = "decfile.lark"
-            with data.basepath.joinpath(filename).open() as f1:
-                self._grammar = f1.read()
+
+        self._load_grammar(filename, parser, lexer, **options)
+
+    def load_additional_decay_models(self, *models: str) -> None:
```
eduardo-rodrigues, line 309: This method requires a docstring, even more because it is a public method.

The code seems to imply that  `models` is only taken into account if `self._additional_decay_models ` is None. That is most probably the default usage. But imagine somebody does provide additional models in the `__init__` method already; then this call will override, right? One should probably expand the test below to confirm and ensure the code does what it is expected in such a case. In fact, one should give examples in https://github.com/scikit-hep/decaylanguage/blob/master/README.md#advanced-usage. Let me know if you have limited time and prefer that we do this as a follow-up, which is OK.

src/decaylanguage/dec/dec.py
```diff
@@ -281,18 +294,80 @@ def load_grammar(
...
         self._grammar_info = dict(
-            lark_file=filename, parser=parser, lexer=lexer, **options
+            lark_file=filename,
+            parser=parser,
+            lexer=lexer,
+            edit_terminals=self._generate_edit_terminals_callback(),
+            **options,
         )
 
+    def _generate_edit_terminals_callback(self) -> Callable[[TerminalDef], None]:
```
eduardo-rodrigues, line 350: Can you also add some docstring here motivating this internal helper? It does help when one reads things back months down the line :-).

src/decaylanguage/dec/dec.py
```diff
@@ -281,18 +294,80 @@ def load_grammar(
...
+        if self._additional_decay_models is None:
+            decay_models = known_decay_models  # type: tuple[str, ...]
+        else:
+            decay_models = tuple(
+                chain.from_iterable([known_decay_models, self._additional_decay_models])
+            )
+
+        def edit_model_name_terminals(t: TerminalDef) -> None:
+            """
+            Edits the terminals of the grammar to replace the model name placeholder with the actual names of the models.
```
eduardo-rodrigues, line 360:
```suggestion
            Edits the terminals of the grammar to replace the model name placeholder with the actual names of the models,
            see `MODEL_NAME_PLACEHOLDER` in the default grammar file decaylanguage/data/decfile.lark.
```

eduardo-rodrigues (changes requested):
Hi. I made some more comments and final suggestions. Let me know how you feel about it and we finalise.

src/decaylanguage/dec/dec.py
sognetic: Ah, thank you for catching this, I'll add a docstring. However, I have to correct the second part (if I have understood you correctly): The method will always work, if `self._additional_decay_models` is already some kind of iterable it will just chain the two iterables together. This way `self._additional_decay_models` can be called multiple ways or one could sub-class and provide them in the `__init__`. I'll add a test to make sure this works, you're right that this should be tested.

eduardo-rodrigues: Many thanks.

README.md
```diff
@@ -134,17 +134,14 @@ from decaylanguage.dec import known_decay_models
...
 
 Say you have to deal with a decay file containing a new model not yet on the list above.
-Running the parser as usual will result in a `UnexpectedToken` exception.
-It is nevertheless easy to deal with this issue; no need to wait for a new release.
-It is just a matter of adding the model name to the list in `decaylanguage/data/decfile.lark`
-(or your private copy), see the line
-`MODEL_NAME.2 : "BaryonPCR"|"BTO3PI_CP"|"BTOSLLALI"|...`,
-and then proceed as usual apart from adding an extra line to call to `load_grammar`
-to specify the Lark grammar to use:
+Running the parser as usual will result in an error in the `ModelAlias` parsing.
```
eduardo-rodrigues, line 137:
```suggestion
Running the parser as usual will result in an error in the model parsing.
```

src/decaylanguage/dec/dec.py
```diff
@@ -136,6 +138,10 @@ def __init__(self, *filenames: str | os.PathLike[str]) -> None:
             Any
         ) = None  # Particle decays found in the decay file
 
+        self._additional_decay_models: None | Iterable[
+            str
+        ] = None  # Additional decay models not (yet) known to decaylanguage
```
eduardo-rodrigues, line 143:
```suggestion
        ] = None  # Additional decay models not (yet) known to DecayLanguage
```

src/decaylanguage/dec/dec.py
```diff
@@ -281,18 +294,97 @@ def load_grammar(
...
+        self._load_grammar(filename, parser, lexer, **options)
+
+    def load_additional_decay_models(self, *models: str) -> None:
+        """
+        Add one or more EvtGen decay models in addition to the ones already provided via
+        `decaylanguage.dec.enums.known_decay_models`.
+
+        Parameters
+        ----------
+        *models: str
```
eduardo-rodrigues, line 316:
```suggestion
        models: str
```

tests/data/test_custom_decay_model.dec
```diff
@@ -0,0 +1,6 @@
+# Simple decay with a custom model name
```
eduardo-rodrigues, line 1:
```suggestion
# Example decfile for testing purposes:
# Simple decay with custom models
```

eduardo-rodrigues (approved):
Thank you again for this great contribution 👍!

sognetic:
Thank you for the great review and for adding the finishing touches!

PR #400, opened 2023-11-28
fix: support passing though dm, nicer typing

henryiii:
The changes from #392, but in a branch that won't get force pushed by pre-commit weekly. Rebased after #398. Closes #399.

### DIFF
--- a/src/decaylanguage/decay/decay.py
+++ b/src/decaylanguage/decay/decay.py
@@ -257,7 +257,7 @@ class DecayMode:
         """
         self.bf = bf
         if daughters is None and "fs" in info:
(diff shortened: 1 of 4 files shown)


src/decaylanguage/decay/decay.py
```diff
@@ -241,6 +256,8 @@ def __init__(
         True
         """
         self.bf = bf
+        if daughters is None and "fs" in info:
+            self.daughters = DaughtersDict(info.pop("fs"))
```
henryiii, line 260:
```suggestion
            daughters = info.pop("fs")
```

PR #401, opened 2023-11-30
Support all available Belle II decay models

eduardo-rodrigues:
FYI @sognetic, this is what I could find in Basf2.


eduardo-rodrigues:
FYI some files still give parsing errors, namely
['basf2/decfiles/dec/1113042002.dec',
 'basf2/decfiles/dec/1263700001.dec',
 'basf2/decfiles/dec/1263752000.dec',
 'basf2/decfiles/dec/3470012000.dec',
 'basf2/decfiles/dec/3470012001.dec',
 'basf2/decfiles/dec/3470030000.dec',
 'basf2/decfiles/dec/3470030001.dec',
 'basf2/decfiles/dec/BSemiTauonic_BSTD.dec',
 'basf2/decfiles/dec/BSemiTauonic_BSTD_2HDMTYPE2.dec']

I need to look into those, and help from Belle-II colleagues would be appreciated.
For example the first file happens to have spurious characters a line after the `End` statement. That should not be a problem per se, meaning one could imagine a parser that ignores anything after the `End` statement, but it still is true that those spurious characters are a mistake.

Let's see for the rest of the files ...

eduardo-rodrigues:
The 2nd file as the keyword `PHOTOS` twice in a decay line. Again, that is not a major issue in itself, and could easily be supported, but I was likely a copy-paste mistake to have `... PHOTOS   PHOTOS ... `.

The 3rd file has a line finishing with ` LLSW 0.71 -1.6 -0.5 2.9;sig`, and it is not allowed to get non-comment characters after the end-of-line `;` (`... LLSW 0.71 -1.6 -0.5 2.9;#sig` would be OK).

Then files 3470012000.dec and 3470012001.dec in fact aren't dec files, since ASCII tables ;-).

Etc. I will stop here.

I reckon I am going to merge and you see what to do with these files. Maybe worth opening an issue in the Basf2 repo and point to this discussion?

HTH.

PR #429, opened 2024-04-02
chore: use ruff format instead of black

henryiii:
This is faster and simpler, doesn't require a separate install.


eduardo-rodrigues (approved):

henryiii:
I missed adding notebooks, can do that in a sec.

eduardo-rodrigues:
> I missed adding notebooks, can do that in a sec.

Advance thanks :).

eduardo-rodrigues (approved):
Feel free to merge when happy.

henryiii:
Added ruff linting to the notebooks too.

PR #18, opened 2019-03-29
Remove leftover notebook

eduardo-rodrigues:

PR #440, opened 2024-05-29
Add support for EvtGen model HypNonLepton

eduardo-rodrigues:
FYI @sognetic, with this update all Belle II files in the current version of basf2 are parsed with no issues aside the one file 1263700001.dec, which contains the line
`0.5   anti-D*0    mu+ nu_mu        PHOTOS         PHOTOS  HQET3 0.912 1.205 1.15 1.404 0.854; #18.09.2018: updated to main DECAY_BELLE2.DEC`
and DecayLanguage does not allow for twice the keyword `PHOTOS` in a decay line - not a bug per se but I would say a buglet. Maybe worth fixing upstream.

### DIFF
--- a/src/decaylanguage/dec/enums.py
+++ b/src/decaylanguage/dec/enums.py
@@ -74,6 +74,7 @@ known_decay_models = (
     "HQET3",
     "HQET2",
     "HQET",
+    "HypNonLepton",
     "ISGW2",
     "ISGW",
     "KNUNU",
--- a/tests/dec/test_known_decay_models.py
+++ b/tests/dec/test_known_decay_models.py
@@ -15,7 +15,6 @@ from decaylanguage.dec.enums import known_decay_models
 DIR = Path(__file__).parent.resolve()
 
 
-# TODO: actually test all models - takes time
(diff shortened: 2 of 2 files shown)


sognetic:
Hi @eduardo-rodrigues , thanks for the notification! I don't have access to upstream anymore but let me ping @GiacomoXT

GiacomoXT:
@eduardo-rodrigues thanks for having spotted it! Indeed having the keyword `PHOTOS` twice is useless, I will fix it upstream in basf2.

eduardo-rodrigues:
Hi both. Glad I could help.

In the future do not hesitate to run the parsing over your files for validation :-).

GiacomoXT:
> Hi both. Glad I could help.
> 
> In the future do not hesitate to run the parsing over your files for validation :-).

Sure, it's a good idea.
 
(Thanks to @sognetic , since a bit of time we have a unit test in place that runs EvtGen with all the new signal decfiles added to basf2 via merge request, so luckily we are not adding anymore broken decfiles! But using DecayLanguage would be definitely faster.)

PR #561, opened 2026-03-15
Feature: minimum_effective_bf in build_decay_chains

vvsagar:
Another important feature to be able to deal with mothers with large and complex decay chains.

I already added some tests and docs, and I look forward to your feedback @eduardo-rodrigues :)


eduardo-rodrigues (approved):
Hello again @vvsagar, and apologies that I have been slow. I did find again my notes from a few years ago with some ideas :). I will try and post a few little things in the enhancement issue ...

I think this is basically ready. There is maybe an extra test worth having to be 100% sure all is good: that is the case when one sets a minimum BF to consider and at the same time sets certain particles to stable, whereby doing so makes at least one decay path that would otherwise fail, now pass because the BF of the particle (decay) set to stable is effectively 1. I am being manic here, but can you add such a test? I reckon it is useful.

eduardo-rodrigues:
I forgot to say - let me know if you would prefer that I add some of these tests. I understand I am asking more than standard :).

vvsagar:
Hi @eduardo-rodrigues,  I added all the tests you suggested and a few more I could think of. Let me know if I missed anything :)

eduardo-rodrigues:
Hey. I was going to say that we can follow this up, as I'm sure you also want to have this really nice update merged. I appreciate the effort. Thank you and talk to you again after the Easter break.

Issue #200, opened 2021-10-28
Support for decay descriptors


admorris:
As part of the [LHCb Ntuple Wizard project](https://lhcb-dpa.web.cern.ch/lhcb-dpa/wp6/ntupling-wizard.html), we have implemented decay descriptor parsing (and rendering) using [pyparsing](https://pypi.org/project/pyparsing/). I would like to port this over to `decaylanguage` and make it easy to add different "grammars" used by different experiments/software packages.

Example grammars:
- LHCb's [`LoKi` decay descriptors](https://twiki.cern.ch/twiki/bin/view/LHCb/FAQ/LoKiNewDecayFinders)
- Belle II's [`DecayDescriptor`](https://software.belle2.org/development/classBelle2_1_1DecayDescriptor.html) class
- [`KFParticle`](https://wiki.bnl.gov/sPHENIX/index.php/KFParticle#The_Decay_Descriptor), used in sPHENIX

We can also think about doing matching and substitution (which we also have functionality for in the Ntuple Wizard) but there it's less clear to me how easy it is to support different conventions from different experiments/packages.

eduardo-rodrigues:
Thanks @admorris for creating this work task / enhancement idea following our chats within LHCb. Thanks also for the links. I was not aware of all of those and this is bound to be interesting and relevant.

You know already that I like the idea of the extension since specifications of decays such as `B0 -> pi+ pi-` make sense and are very much used in Flavour Physics experiments. We should try and provide at least a basis "for everyone", as generic as possible. Once we integrate with the rest then all package functionality can be exploited for free ...

On the spot I would say that it would make sense to draft something say as a `descriptor/` submodule under
https://github.com/scikit-hep/decaylanguage/tree/master/src/decaylanguage, in the same way that we have `dec/` to deal with .dec decay files and `decay` to define decay chain classes.

@henryiii, I hope you also like these ideas :-).

We should try and see if Belle-II colleagues could be interested in this. Let me start by pinging usual suspects such as @daritter and @GiacomoXT (feel free to ping other colleagues).

eduardo-rodrigues:
Hi @admorris, shall we try and move forward with this nice addition? Such a module will be super useful also for run-3 DaVinci, hence I'm very keen. Let me know whether you will have a bit of time for this. I'm happy to help.

admorris:
In #331 I add a function that writes "LHCb-style" decay descriptors e.g. `D*+ -> (D0 -> K- pi+) pi+`

[...]

PR #573, opened 2026-04-27
Add support for parsing decay descriptors

admorris:
Implements the rest of #200 

Added `DecayChain.from_string` method

Decay descriptors are parsed with Lark. A `Transformer` class converts them into `DecayChainDict` objects, which are then used to initialise `DecayChain` objects.

Custom descriptor formats can be used by pointing to another `.lark` file in an argument of `DecayChain.from_string`. These pretty much only have the freedom to modify `ARROW`, `LPAR` and `RPAR`. The rest of the structure is assumed by the Transformer. **i.e.** I did not find a way to support sub-decays written with the mother outside of braces like `A -> B (-> C D) E`

**One glaring limitation** (which is inherent to `DecayChain`/`_build_decay_modes`) is that sub-decays of identically named particles are not supported: e.g.. `"B_s0 -> (phi -> K+ K-) (phi -> K+ K-)"` will result in an exception. This could possibly be handled by adding internal/hidden uniqueness when duplicates are encountered.


eduardo-rodrigues:
Hi @admorris, I am not forgetting to check this. It's just that I have been working on urgent and important suff. Will get back to you very soon.

src/decaylanguage/data/descriptor.lark
```diff
@@ -0,0 +1,29 @@
...
+sub_decay: LPAR decay RPAR
+
+// Terminals
+ARROW: "->"
+LPAR: "("
+RPAR: ")"
+
+// Particle names start with alphanumeric/underscore and can then include
+// common descriptor suffix symbols such as +, -, *, ', and ~.
+PARTICLE: /[A-Za-z0-9_][A-Za-z0-9_+*'~̄-]*/
```
eduardo-rodrigues: This definition does not match that in https://github.com/scikit-hep/decaylanguage/blob/main/src/decaylanguage/data/decfile.lark. Some name may be overlooked, such as those with parentheses?

src/decaylanguage/data/descriptor.lark
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2018-2026, Eduardo Rodrigues and Henry Schreiner.
```
eduardo-rodrigues, line 1: Can you update the README at https://github.com/scikit-hep/decaylanguage/tree/main/src/decaylanguage/data since this one will be public-facing? Just so people can more easily be aware of it and the possibility to change the descriptor definition.

src/decaylanguage/decay/decay.py
```diff
@@ -729,6 +743,60 @@ def _expand_decay_modes(
...
+        return typing.cast(str | DecayChainDict, items[0])
+
+    def daughters(self, items: list[Any]) -> list[str | DecayChainDict]:
+        # daughters: daughter+
+        return [item for item in items if isinstance(item, (str, dict))]
+
+    def sub_decay(self, items: list[Any]) -> DecayChainDict:
+        # sub_decay: LPAR decay RPAR
+        for item in items:
+            if isinstance(item, dict):
```
eduardo-rodrigues, line 814: Just for my understanding - this code here would only ever return the first found dict in the list. Is there some subtlety I'm missing for things to work overall, likely thanks to the way the Lark Transformer works?

src/decaylanguage/decay/decay.py
```diff
@@ -795,6 +863,73 @@ def from_dict(cls, decay_chain_dict: DecayChainDict) -> Self:
...
+
+        Parameters
+        ----------
+        descriptor : str
+            The decay descriptor string, e.g.
+            ``"D*+ -> (D0 -> K+ pi-) pi+"``.
+        grammar_file : str or Path, optional
+            Path to a custom Lark grammar file for descriptor parsing.
+            If not provided, the default grammar is used.
+        Returns
```
eduardo-rodrigues, line 941:
```suggestion

        Returns
```

src/decaylanguage/decay/decay.py
```diff
@@ -795,6 +863,73 @@ def from_dict(cls, decay_chain_dict: DecayChainDict) -> Self:
 
         return cls(mother, decay_modes)
 
+    @classmethod
+    def from_string(
```
eduardo-rodrigues, line 923: At some point it would make sense to "synchronise" this `from_string` function with the existing `to_string` one, since they should effectively be the "mirror of each other". Else one would name this function to `from_descriptor`. WDYT?

src/decaylanguage/decay/decay.py
```diff
@@ -795,6 +863,73 @@ def from_dict(cls, decay_chain_dict: DecayChainDict) -> Self:
...
+    ) -> Self:
+        """
+        Construct a ``DecayChain`` by parsing a descriptor string.
+
+        Parameters
+        ----------
+        descriptor : str
+            The decay descriptor string, e.g.
+            ``"D*+ -> (D0 -> K+ pi-) pi+"``.
+        grammar_file : str or Path, optional
```
eduardo-rodrigues, line 937: This could be a place to state where the default grammar is available? Seems useful and relevant. This would be similar to the docs of `edit_model_name_terminals` in https://github.com/scikit-hep/decaylanguage/blob/main/src/decaylanguage/dec/dec.py. WDYT?

src/decaylanguage/decay/decay.py
```diff
@@ -795,6 +863,73 @@ def from_dict(cls, decay_chain_dict: DecayChainDict) -> Self:
 
         return cls(mother, decay_modes)
 
+    @classmethod
+    def from_string(
+        cls,
+        descriptor: str,
+        *,
+        grammar_file: str | Path | None = None,
```
eduardo-rodrigues, line 927: You could do similary to https://github.com/scikit-hep/decaylanguage/blob/main/src/decaylanguage/dec/dec.py#L300, meaning provide the name of the default grammar, stating where it is located (I realise the code docstring I am refering to could be improved a bit). Similarly, you could use https://github.com/scikit-hep/decaylanguage/blob/main/src/decaylanguage/dec/dec.py#L313 below?

src/decaylanguage/decay/decay.py
```diff
@@ -795,6 +863,73 @@ def from_dict(cls, decay_chain_dict: DecayChainDict) -> Self:
...
+
+        if not grammar_file.exists():
+            raise FileNotFoundError(f"Grammar file not found: {grammar_file}")
+
+        with grammar_file.open() as f:
+            grammar = f.read()
+
+        # Parse the descriptor
+        try:
+            parser = Lark(grammar, parser="lalr", transformer=None)
```
eduardo-rodrigues: Add also the default `lexer="auto"`?

tests/data/descriptor_alt.lark
```diff
@@ -0,0 +1,29 @@
...
+sub_decay: LPAR decay RPAR
+
+// Terminals
+ARROW: "=>"
+LPAR: "{"
+RPAR: "}"
+
+// Particle names start with alphanumeric/underscore and can then include
+// common descriptor suffix symbols such as +, -, *, ', and ~.
+PARTICLE: /[A-Za-z0-9_][A-Za-z0-9_+*'~̄-]*/
```
eduardo-rodrigues, line 28: Ditto

eduardo-rodrigues: In any case, why do you need this copy? Then worth having a comment about it in the file, I reckon.

eduardo-rodrigues: I see you do test with a double arrow. Yeah, just add a comment about it to be trivial to any reader :).

eduardo-rodrigues (commented):
Thank you so much for this, @admorris! It's a really nice enhancement 👍.

I left a few little suggestions but this is looking excellent anway.

I am well aware of the limitation you point out. It does annoy me.

src/decaylanguage/decay/decay.py
admorris: The `items` parameter is always a list. In this case we should always expect a list of length 1. Maybe I could raise an exception if it's a different lenght.

src/decaylanguage/data/descriptor.lark
admorris: Allowing parentheses is causing a headache where it matches the parentheses from a sub-decay as part of the first or last particle name. I am trying to debug it without coming up with grotesque regex patterns.

src/decaylanguage/decay/decay.py
eduardo-rodrigues: Seems better IMO. Basically anything that is not in the expected format should get an exception.

src/decaylanguage/decay/decay.py
eduardo-rodrigues: I see you dealt with this in your most recent commits. Resolving.

src/decaylanguage/data/descriptor.lark
eduardo-rodrigues: Ah, indeed that's not an easy one!

src/decaylanguage/data/descriptor.lark
eduardo-rodrigues: I see you dealt with this in your most recent commits. Resolving.

eduardo-rodrigues:
Hey @admorris, let me know your thoughts on this enhancement and whether you would prefer to have a follow-up for the limitations and some of the matters discussed above.

BTW, for the issue with parsing parenthesis for particle names. One thing that differentiates parentheses for particle names wrt parenthesis denoting decays is that the latter always contain an arrow inside, so between ( and ).

src/decaylanguage/decay/decay.py
admorris: Done in aa68c2a

src/decaylanguage/decay/decay.py
admorris: Like this? aa68c2a

tests/data/descriptor_alt.lark
admorris: Done in 61d5892

src/decaylanguage/decay/decay.py
admorris: The pytest `test_from_string_to_string` demonstrates they mirror eachother in the specific case of using the default grammar.

What if this function is renamed `from_descriptor`, then `from_string` just invokes `from_descriptor` with the default grammar?

src/decaylanguage/decay/decay.py
eduardo-rodrigues: Yep, thanks.

src/decaylanguage/decay/decay.py
eduardo-rodrigues: Thanks for adding the exception raise.

src/decaylanguage/decay/decay.py
eduardo-rodrigues: Yes, that sounds good to me.

eduardo-rodrigues:
Hey. So aside the trivial comment above, I think there are only 2 things to get sorted:$
- Add some more info/doc to the descriptor.lark file as it got reasonably complicated to deal with particle names involving parentheses.
- Enhance the tests to include a couple of complicated particle names with parentheses, so names of the sort psi(2S) or Upsilon_3(1D) or anti-Lambda_b(5920)0.

eduardo-rodrigues:
Hi @admorris, you may have seen the major improvements made to the package recently, thanks to @henryiii. We are moving towards a 1.0 release ...

Let me know if you will be able to pick up the work here after a rebase.

FYI you can now do the following:
```
In [2]: dm_Bs = DecayMode(1, 'phi phi')
   ...: dm_phi = DecayMode(1, 'K+ K-')
   ...: dc = DecayChain('Bs', {'Bs':dm_Bs, 'phi':dm_phi})

In [3]: dc_dict = dc.to_dict()

In [4]: DecayChain.from_dict(dc_dict).to_dict() == dc_dict
Out[4]: True

In [5]: dc.to_string()
Out[5]: 'Bs -> (phi -> K+ K-) (phi -> K+ K-)'
```

admorris:
Very good! I will have a look

henryiii:
Whenever you are ready, I can do an Opus 4.8 or GPT 5.5 review of the PR if that sounds helpful. (if you are using one of those, or already ran a review with one, let me know and I'll do the other one, otherwise I'll do Opus because I've got higher limits on that)

eduardo-rodrigues:
> Whenever you are ready, I can do an Opus 4.8 or GPT 5.5 review of the PR if that sounds helpful. (if you are using one of those, or already ran a review with one, let me know and I'll do the other one, otherwise I'll do Opus because I've got higher limits on that)

Hello @henryiii, I will get back to this PR now that version 1.0 is out. Before I do a "physics" review, if you can then run an Opus or similar review, that will be great. Thank you.

eduardo-rodrigues:
> > Whenever you are ready, I can do an Opus 4.8 or GPT 5.5 review of the PR if that sounds helpful. (if you are using one of those, or already ran a review with one, let me know and I'll do the other one, otherwise I'll do Opus because I've got higher limits on that)
> 
> Hello @henryiii, I will get back to this PR now that version 1.0 is out. Before I do a "physics" review, if you can then run an Opus or similar review, that will be great. Thank you.

Hi @henryiii, a lot has been happening recently. I would like to get back to this now that the other developments are available in the latest 1.x releases. Very happy, then, to have an AI review from you. Let me know, thanks.

henryiii:
Claude Fable standard review:

# Review of PR #573 — Add support for parsing decay descriptors

## Overview

Adds `DecayChain.from_string`, the inverse of `to_string`, closing out #200. A new `data/descriptor.lark` grammar parses descriptors like `"D*+ -> (D0 -> K+ pi-) pi+"`; a Lark `Transformer` (`_DescriptorTreeToDict`) converts the parse tree into a `DecayChainDict` fed to the existing `from_dict`. Custom grammars can be supplied via `grammar_file`. `__eq__`/`__hash__` are added to `DecayMode` and `DecayChain` to support the equality-based tests.

The design fits the codebase well — same Lark-grammar-in-`data/` pattern as `dec/` and `modeling/`, and reusing `from_dict` keeps the new surface small. I verified the grammar handles the tricky cases correctly: `K*(892)0` (parenthesized mass suffix, with or without a sub-decay), flat descriptors with no sub-decay, and trailing suffixes all parse right. The explicit whitespace handling (instead of `%ignore`) is justified — it keeps `K_1 (1270)` from gluing into one particle.

## Issues

**1. Error contract is leaky for same-name sub-decays (verified by running the PR code).** The docstring promises `ValueError` for malformed input, but only `LarkError` and transformer exceptions are wrapped; whatever `from_dict` raises escapes as-is. I tested:

- `"B_s0 -> (phi -> K+ K-) (phi -> K+ K-)"` (identical duplicates) actually **works fine**, contrary to the limitation stated in the PR description.
- `"B_s0 -> (phi -> K+ K-) (phi -> pi+ pi- pi0)"` (same name, *different* modes) raises `RuntimeError: Input is not a single decay chain!` — a confusing message that doesn't tell the user the real problem, and the wrong exception type per the docstring.

Suggest wrapping the `from_dict` call too, re-raising as `ValueError` with a message naming the duplicated particle, and documenting the limitation in the `from_string` docstring rather than only in the PR body.

**2. Misleading error for a broken custom grammar** (`decay.py:973`). `Lark(grammar, ...)` is constructed inside the `try` that catches `LarkError`, so a syntax error in a user's custom grammar file is reported as `"Failed to parse descriptor '...'"`. Construct the parser outside that `try` (or catch `GrammarError` separately). Also worth documenting that custom grammars must be LALR-compatible, and that a bad `grammar_file` path raises `FileNotFoundError`.

**3. Parser is rebuilt on every call.** LALR table construction runs per `from_string` invocation. Anyone parsing descriptors in a loop pays it repeatedly. A module-level `@functools.lru_cache` keyed on the grammar text (or path) is a two-line fix.

**4. RST underline too short** (`data/README.rst`). The ``` ``descriptor.lark`` ``` title is 19 chars but the underline is 18 dashes (verified) — docutils emits a "title underline too short" warning.

**5. Missing `CHANGELOG.md` entry.** Project convention per CLAUDE.md, and doubly warranted here: adding `__eq__` to `DecayMode`/`DecayChain` changes public behavior (equality was identity-based before), which users should hear about in release notes, not just the new `from_string` feature.

## Test coverage

The positive-path tests are good — repurposing the existing formatter examples for round-trip tests is exactly right. Gaps:

- No negative tests: nothing asserts `ValueError` for malformed descriptors (`"D0 ->"`, empty string, unbalanced parens).
- No flat-descriptor test (`"D0 -> K- pi+"` — no sub-decay). It works; a test would keep it working.
- No test for the same-name sub-decay cases from issue 1, in either the working or failing variant.
- No test for a suffix particle like `K*(892)0`, which is the case the `pgroup_paren` machinery exists for — currently that whole code path is untested.

## Minor / style

- `Lark(grammar, parser="lalr", transformer=None, lexer="auto")`: `transformer=None` and `lexer="auto"` are the defaults — drop them. (Alternatively, passing the transformer directly to a LALR `Lark` is Lark's intended fast path, though the current two-step form gives cleaner error separation.)
- `__hash__` raising `TypeError` works, but the conventional `__hash__ = None` also makes `isinstance(x, collections.abc.Hashable)` return `False`; the current form leaves instances looking hashable to ABC checks. If it was done to satisfy strict mypy, a brief comment would save the next reader the puzzle.
- `psuffix` returning `""` (`decay.py` transformer): the fallback is unreachable — `psuffix` always has exactly one non-ws child; `return self._item_text(items[0])` would be more honest.
- `grammar_file` handling: `Path(grammar_file).read_text(...)` covers both `str` and `Path` without the `isinstance` branch.
- The alternative test grammar's `PARTICLE` regex contains an invisible combining macron (U+0304) in its character class — presumably intentional for `p̄`-style names, but worth a comment since it's indistinguishable from a stray character.

## Verdict

Solid, well-scoped feature that follows the project's grammar-driven architecture, and the grammar itself held up to every edge case I threw at it. The main asks before merge: tighten the error contract for same-name sub-decays (issue 1 — the actual behavior differs from what the PR description claims, in both directions), fix the misleading custom-grammar error path, add the missing negative/edge tests, and the changelog entry plus one-character RST fix.

eduardo-rodrigues:
Super, thank you much, @henryiii. There's now some bits of work to do :).

tests/decay/test_descriptor.py
```diff
@@ -21,27 +24,76 @@
...
+    assert result == expected
+
+
+@pytest.mark.parametrize(
+    "descriptor",
+    [
+        "D0 -> (K_S0 -> pi+ pi-) (pi0 -> gamma gamma)",
+        "D*+ -> (D0 -> (K_S0 -> pi+ pi-) (pi0 -> gamma gamma)) pi+",
+        "B0 -> (D- -> K+ pi- pi-) (tau+ -> anti-nu_tau pi+ pi+ pi-) nu_tau",
+        "B_s0 -> (phi -> K+ K-) (phi' -> pi+ pi- pi0)",
```
eduardo-rodrigues, line 92: Add here `"B_s0 -> (phi -> K+ K-) (phi -> K+ K-)"` as per my comment above on the limitation being a thing of the past.

eduardo-rodrigues:
Hello @admorris, cc @henryiii. The review is very good. There are things that we could leave for a follow-up enhancement. Other small things like better coverage would be useful, though. If you're really low on time let me know and I can try and step in in a follow-up - would avoid trying to work on your fork, which would require a sync with main.

tests/decay/test_descriptor.py
```diff
@@ -21,27 +24,76 @@
...
+    ],
```
eduardo-rodrigues, line 93:
```suggestion
        "B_s0 -> (phi -> K+ K-) (phi -> K+ K-)",
    ],
```

tests/decay/test_descriptor.py
```diff
@@ -21,27 +24,76 @@
...
+        grammar_file=Path(__file__).parent.parent / "data" / "descriptor_alt.lark",
+    )
+    expected = DecayChain.from_string(conventional)
+    assert result == expected
+
+
+@pytest.mark.parametrize(
+    "descriptor",
+    [
+        "D0 -> (K_S0 -> pi+ pi-) (pi0 -> gamma gamma)",
```
eduardo-rodrigues, line 89:
```suggestion
        "D0 -> K- pi+",
        "D0 -> (K_S0 -> pi+ pi-) (pi0 -> gamma gamma)",
```

tests/decay/test_descriptor.py
```diff
@@ -21,27 +24,78 @@
...
+
+
+@pytest.mark.parametrize(
+    "descriptor",
+    [
+        "D0 -> K- pi+",
+        "D0 -> (K_S0 -> pi+ pi-) (pi0 -> gamma gamma)",
+        "D*+ -> (D0 -> (K_S0 -> pi+ pi-) (pi0 -> gamma gamma)) pi+",
+        "B0 -> (D- -> K+ pi- pi-) (tau+ -> anti-nu_tau pi+ pi+ pi-) nu_tau",
+        "B_s0 -> (phi -> K+ K-) (phi' -> pi+ pi- pi0)",
```
eduardo-rodrigues, line 93:
```suggestion
        "B_s0 -> (phi -> K+ K-) (phi' -> pi+ pi- pi0)",
        "B_s0 -> (phi -> K+ K-) (phi -> K+ K-)",
```

Issue #581, opened 2026-06-11
Claude Fable review for decaylanguage


henryiii:
:robot: _AI text below_ :robot:

All high-severity claims I spot-checked reproduce. Here's the consolidated review.

# Review: decaylanguage

I ran four parallel reviewers over `dec/`, `decay/`, `modeling/`, and the packaging/CI config, then independently verified the most serious claims. The core `dec`/`decay` code is in good shape; the real problems concentrate in `modeling/` and a couple of long-standing edge cases.

## High-severity bugs (all reproduced)

**`ModelAlias` used in more than one decay block crashes the parser** — `dec.py:214` builds one shared subtree per model alias, and `DecayModelAliasReplacement._replacement` (dec.py:1168) returns that *same* children list for every use. `DecayModelParamValueReplacement` then mutates the shared tokens in place once per containing tree; on the second visit the value is already a float and `t.value[0]` (dec.py:1227) raises `TypeError: 'float' object is not subscriptable`. I reproduced this with a minimal dec file using `Define` + `ModelAlias` in two `Decay` blocks. Fix: deepcopy inside `_replacement` (and the redundant outer deepcopy at line 214 can then go).

**`DecayChain.from_dict` can't read back its own `to_dict` output** when the same decaying particle appears twice in a final state — `decay.py:544` raises `RuntimeError("Input is not a single decay chain!")` on the second identical embedded sub-dict. Reproduced with `eta -> pi0 pi0`, `pi0 -> gamma gamma`. The guard should allow an identical repeated mode.

**`modeling/` has a cluster of output-correctness bugs**, two of which are baked into the reference test output `tests/output/DtoKpipipi_v2.cu`:
- The amplitude fix/free flag is **inverted** end-to-end: `ampgentransform.py:41-46` returns is-fixed, but `cplx_decay_line` treats it as is-free — AmpGen `Fix=2` lines come out as free GooFit variables and vice versa.
- Polar→cartesian error propagation (`amplitudechain.py:109-113`) uses `sin(dθ)`/`cos(dθ)` instead of `sin(θ)·dθ`/`cos(θ)·dθ`, so zero input errors produce an imaginary error of 1.0 (visible in the reference output).
[...]

PR #582, opened 2026-06-11
fix(utils): fix split() trailing-comma bug, single-pass filter_lines, remove dead code, unbounded cache

henryiii:

Part of #581

## Summary

- **`split()` trailing-comma bug**: the `i+1==len(x)` early-exit ran before the comma check, so `split("a,")` returned `["a,"]` instead of `["a", ""]`. Rewrote as a single linear scan tracking bracket depth, eliminating the O(n²) repeated `x = x[i+1:]` string slicing.
- **`filter_lines()` double-match**: collected matched lines and residuals in one pass instead of calling `matcher.match()` twice per line.
- **Dead code in `_from_group_dict_list`**: removed `mat.get("prime")` branch — the `_getname` regex has no `prime` capture group so the condition could never be true.
- **`lru_cache` sizing**: bumped `maxsize=64` to `maxsize=None` on `charge_conjugate_name`, which is a pure `str→str` mapping called with hundreds of distinct species in typical decay files.

## Test plan

- [ ] Added parametrised tests for `split()` covering nesting, trailing comma, leading comma, bare comma, empty string, and deeply nested brackets.
- [ ] Added parametrised tests for `filter_lines()` covering normal matches, empty input, and no-match input.
- [ ] `uv run pytest` — 312 passed, 1 skipped (pre-existing unconditional skip).
- [ ] `prek -a --quiet` — clean.

### DIFF
--- a/src/decaylanguage/utils/particleutils.py
+++ b/src/decaylanguage/utils/particleutils.py
@@ -19,7 +19,7 @@ from particle.converters import (
 from particle.exceptions import MatchingIDNotFound
 from particle.particle.enums import Charge_mapping
 
-cacher = lru_cache(maxsize=64)
+cacher = lru_cache(maxsize=None)
 
 
 @cacher
@@ -139,9 +139,6 @@ def _from_group_dict_list(mat: dict[str, Any]) -> list[Particle]:
         name += f"({mat['family']})"
     if mat["state"]:
         name += f"({mat['state']})"
-
-    if mat.get("prime"):
-        name += "'"
 
     if mat["star"]:
         name += "*"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@
   - Migrated `ModelDecay` and `AmplitudeChain` from old-style `@attr.s`/`attr.ib` to modern `@attrs.define`/`attrs.field`.
   - Removed dead `graphviz` import guard in `decay.py` (graphviz is a hard dependency).
   - Various fixes and minor improvements.
+* Utilities:
+  - Fixed a bug in `split()` where a trailing comma was not handled correctly (e.g. `split("a,")` returned `["a,"]` instead of `["a", ""]`); rewrote as a single linear scan to eliminate O(n²) repeated string slicing.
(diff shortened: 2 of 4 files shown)


src/decaylanguage/utils/particleutils.py
```diff
@@ -140,9 +140,6 @@ def _from_group_dict_list(mat: dict[str, Any]) -> list[Particle]:
     if mat["state"]:
         name += f"({mat['state']})"
 
-    if mat.get("prime"):
-        name += "'"
-
```
henryiii, line 145: I wonder if this was supposed to be in the original regex.

eduardo-rodrigues: Quickly as I'm totally taken by something else at the moment and cannot review properly - I even wonder if this helper function is still needed. As for primes "'", some particles do have it in the name, a prominent example being
```
>>> etap_958
<Particle: name="eta'(958)", pdgid=331, mass=957.78 ± 0.06 MeV>
```
Other charmed baryons also have a "prime partner".

henryiii: This is doing the ampgen -> PDG name conversion, it was removed because the ampgen regex doesn't include a prime group. I'm pretty sure it does, though. The current change doesn't affect anything, so it's fine, but maybe this is a fix in the future. Probably should fuzz on name conversions if this matters.

I'm biased to just merge any "fixes" proposed toward the ampgen stuff, and then later see if I end up working on it again, or just remove it entirely eventually.

eduardo-rodrigues: Hi @henryiii, I see you have undrafted this, hence reviewed ... but, completely? Even here?

I am going to resolve and review this myself. Let me know if you are against merging, else I will go ahead when done and happy.

eduardo-rodrigues (approved):
LGTM

src/decaylanguage/utils/particleutils.py
eduardo-rodrigues: Let's go with your approach.

PR #583, opened 2026-06-11
fix(dec): fix several bugs in the .dec file parser

henryiii:

Part of #581

This PR fixes a set of bugs in the `dec/` subsystem (`src/decaylanguage/dec/dec.py`), with regression tests for each.

## Fixes

1. **ModelAlias shared-subtree crash (high).** A `ModelAlias` used in more than one decay returned the same children list for every use, so all decay trees shared the same `Token` objects. `DecayModelParamValueReplacement` then mutated those tokens in place once per containing tree, and the second visit crashed with `TypeError: 'float' object is not subscriptable`. Fixed by deep-copying each subtree at replacement time inside `DecayModelAliasReplacement._replacement`, and dropping the now-redundant outer `deepcopy` in `parse`.

2. **`print_decay_modes(ascending=...)` broken (medium).** The method always sorted descending and ignored `ascending`; the `scale` normalization used `i = -1 if ascending else 0`, which combined with the descending sort picked the *smallest* BF. Now sorts according to `ascending` and always normalizes `scale` by the actual largest BF.

3. **Missing f-string prefix (low).** The `scale` out-of-range error message lacked the `f` prefix, so `{scale}` was printed literally.

4. **`from_string` vs `__init__` inconsistency (low).** `__init__` strips intermediate lines starting with `End` (but not `Enddecay`); `from_string` did not, so identical content could parse from a file but fail from a string. Extracted the filtering into a shared `_filter_lines` helper used by both.

5. **`ChargeConjugateReplacement` mutated the caller's dict (low).** The user-provided `charge_conj_defs` dict was used as a cache and silently polluted, including caching `ChargeConj(...)` failure sentinels that the reverse lookup could then match. Now copies the dict internally and does not cache failure sentinels.

6. **`get_lineshape_settings` masked its own diagnostics (low).** The specific "redefined"/structure `RuntimeError`s raised inside the `try` were swallowed by the function's blanket `except Exception` and re-wrapped generically; the inner `from None` also contradicted the outer `from err`. The specific `RuntimeError`s now propagate, and the contradictory `raise ... from` usage was cleaned up.

7. **Dead `except RuntimeError` in `get_branching_fraction` (low).** Malformed input raises `AttributeError`/`IndexError`/`ValueError`, never `RuntimeError`, so the friendly re-raise never fired. Now catches the actual exception types.

8. **`load_additional_decay_models` (low).** It was a silent no-op if the grammar was already loaded, and stored a one-shot `itertools.chain` iterator that was exhausted on first materialization. Now warns when called after the grammar is loaded and stores a materialized list.

## Testing

- `uv run pytest`: 307 passed, 1 skipped.
- `prek -a --quiet`: clean.
- Added regression tests in `tests/dec/test_issues.py` and `tests/dec/test_dec.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+* Parsing of decay files (aka .dec files):
+  - Fixed a crash when a `ModelAlias` is used in more than one decay, caused by sharing the same parse-tree tokens across decays (the model-parameter substitution mutated them in place).
+  - Fixed `DecFileParser.print_decay_modes(ascending=...)`, which previously ignored the flag and always sorted descending; `scale` now always normalizes to the largest branching fraction.
+  - Made `DecFileParser.from_string` filter intermediate `End` lines like the file-based constructor, so identical content parses the same way either way.
+  - `ChargeConjugateReplacement` no longer mutates the caller's `charge_conj_defs` dict and no longer caches `ChargeConj(...)` failure sentinels.
+  - `get_lineshape_settings` no longer masks its own specific "redefined"/structure diagnostics behind a generic error.
+  - `get_branching_fraction` now catches the exception types actually raised by malformed input.
+  - `load_additional_decay_models` now warns when called after the grammar is loaded and stores a materialized container instead of a one-shot iterator.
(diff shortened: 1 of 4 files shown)


eduardo-rodrigues (approved):
Great catches and improvements 👍!

PR #584, opened 2026-06-11
fix(decay): DecayChain round-trip, flatten, and viewer bugs

henryiii:

Part of #581.

Bug fixes in the `decay/` subsystem (`decay.py` + `viewer.py`), each with a regression test.

### `src/decaylanguage/decay/decay.py`

1. **`DecayChain.from_dict` could not read back its own `to_dict` output (high).** `_build_decay_modes` raised `RuntimeError("Input is not a single decay chain!")` whenever a mother already existed in `decay_modes`. A chain where the same decaying particle appears twice in a final state (e.g. `eta -> pi0 pi0` with `pi0 -> gamma gamma`) embeds the identical sub-dict twice, so round-tripping failed. Now an identical repeated mode is accepted (compared via `to_dict`); a genuinely conflicting redefinition still raises. Also dropped the meaningless `from None`.
2. **`flatten()` crashed when the mother was in `stable_particles` (low).** `keys.index(self.mother)` raised a bare `ValueError` because the mother had been filtered out. The mother is now always decayed (flattening its own decay is the point), even if listed as stable; documented in the docstring.
3. **`stable_particles` did accidental substring matching (low).** The annotation included `str`, so `k not in stable_particles` became substring containment. Changed to `collections.abc.Collection[str]`; behavior remains correct for sets/lists/tuples.
4. **`DaughtersDict` overrode `__add__` but not `__sub__` (low).** `dd1 - dd2` returned a plain `Counter`. Added a matching `__sub__` returning `DaughtersDict`, with a doctest.
5. **Cache-key consistency (trivial).** `charge_conjugate_name(p, pdg_name)` is now called with `pdg_name=` as a keyword to match other call sites and avoid doubling `lru_cache` misses.

### `src/decaylanguage/decay/viewer.py`

6. **Non-reproducible graph output (medium).** The node counter was a module-level global shared across all `DecayChainViewer` instances, so rendering the same chain twice gave different DOT (`dec0` vs `dec1`). The counter is now per-instance (new `_counter` slot; dropped the redundant `iter()` around `count()`).
7. **Unescaped fallback in `safe_html_name` (low).** The `except` branch inserted the raw name into the HTML-like label; names containing `&`, `<`, `>` (legal in `.dec` Alias statements) produced invalid DOT. The fallback now uses `html.escape`.
8. **`_get_graph_defaults` leaked Digraph constructor kwargs into DOT graph attributes (low).** `engine`/`format`/`name` are constructor args, not graph attributes, but appeared in the DOT body. `_get_graph_defaults` now returns only real graph attributes (`rankdir=LR`); the constructor still receives its kwargs via `_get_default_arguments`.

### Tests

- `tests/decay/test_decay.py`: added `__sub__`, `from_dict` repeated-subdecay round-trip, flatten-with-mother-stable, and no-substring-match tests.
- `tests/decay/test_viewer.py`: added reproducibility and no-constructor-kwargs-in-graph-attrs tests; updated `test_simple_decay_chain` node numbers (they previously encoded the shared-global-counter bug).

### Verification

- `uv run pytest`: 305 passed, 1 skipped.
- `prek -a --quiet`: clean.

### DIFF
--- a/tests/decay/test_decay.py
+++ b/tests/decay/test_decay.py
@@ -325,7 +325,7 @@ def test_DecayChain_constructor_from_dict():
 def test_DecayChain_from_dict_roundtrip_repeated_subdecay():
     # A chain where the same decaying particle appears twice in a final state
     # (eta -> pi0 pi0, pi0 -> gamma gamma) embeds the identical sub-dict twice.
-    # from_dict must accept its own to_dict output (regression).
+    # from_dict must accept its own to_dict output.
     dm_eta = DecayMode(1.0, "pi0 pi0")
     dm_pi0 = DecayMode(0.98823, "gamma gamma")
     dc = DecayChain("eta", {"eta": dm_eta, "pi0": dm_pi0})
@@ -464,9 +464,8 @@ def test_DecayChain_flatten_mother_in_stable_particles():
 
 
 def test_DecayChain_flatten_stable_particles_no_substring_match():
-    # stable_particles must not do substring matching: 'pi' should not make
-    # 'pi0'/'pi+' stable (regression for the str annotation accepting a plain
-    # string and falling into substring containment).
+    # stable_particles must not do substring matching:
+    # 'pi' should not make 'pi0'/'pi+' stable.
     dm1 = DecayMode(0.0124, "K_S0 pi0")
     dm2 = DecayMode(0.692, "pi+ pi-")
     dm3 = DecayMode(0.98823, "gamma gamma")
--- a/tests/decay/test_decay.py
+++ b/tests/decay/test_decay.py
@@ -60,6 +60,14 @@ def test_DaughtersDict_add():
     assert len(dd3) == 8
 
 
+def test_DaughtersDict_sub():
+    dd1 = DaughtersDict({"K+": 1, "K-": 2, "pi0": 3})
+    dd2 = DaughtersDict({"K-": 1, "pi0": 1})
+    dd3 = dd1 - dd2
+    assert isinstance(dd3, DaughtersDict)
+    assert dd3 == {"K+": 1, "K-": 1, "pi0": 2}
+
+
 def test_DaughtersDict_to_string():
     dd1 = DaughtersDict({"K+": 1, "K-": 2, "pi0": 3})
     assert dd1.to_string() == "K+ K- K- pi0 pi0 pi0"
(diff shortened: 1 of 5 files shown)


tests/decay/test_decay.py
```diff
@@ -314,6 +322,17 @@ def test_DecayChain_constructor_from_dict():
     assert DecayChain.from_dict(dc_dict).to_dict() == dc_dict
 
 
+def test_DecayChain_from_dict_roundtrip_repeated_subdecay():
+    # A chain where the same decaying particle appears twice in a final state
+    # (eta -> pi0 pi0, pi0 -> gamma gamma) embeds the identical sub-dict twice.
+    # from_dict must accept its own to_dict output (regression).
```
eduardo-rodrigues:
```suggestion
    # from_dict must accept its own to_dict output.
```

tests/decay/test_viewer.py
```diff
@@ -81,6 +81,39 @@ def test_duplicate_arrows(decfilepath, signal_mother, dup):
     assert len(set(ls)) == len(ls)
 
 
+def test_reproducible_node_numbering():
+    # The node counter must be per-instance so rendering the same chain
+    # twice yields identical DOT (regression: it used to be a shared global).
```
eduardo-rodrigues:
```suggestion
    # yields identical DOT.
```

tests/decay/test_viewer.py
```diff
@@ -81,6 +81,39 @@ def test_duplicate_arrows(decfilepath, signal_mother, dup):
     assert len(set(ls)) == len(ls)
 
 
+def test_reproducible_node_numbering():
+    # The node counter must be per-instance so rendering the same chain
```
eduardo-rodrigues:
```suggestion
    # The node counter must be per-instance so rendering the same chain twice
```

tests/decay/test_decay.py
```diff
@@ -431,6 +450,33 @@ def test_DecayChain_flatten_with_stable_particles():
...
+    dc = DecayChain("D0", {"D0": dm1, "K_S0": dm2, "pi0": dm3})
+    dc_flatten = dc.flatten(stable_particles=["D0", "pi0"])
+    assert dc_flatten.mother == "D0"
+    # K_S0 is still decayed, pi0 stays stable, D0 (mother) is decayed too.
+    assert dc_flatten.decays["D0"].daughters == DaughtersDict(["pi+", "pi-", "pi0"])
+
+
+def test_DecayChain_flatten_stable_particles_no_substring_match():
+    # stable_particles must not do substring matching: 'pi' should not make
+    # 'pi0'/'pi+' stable (regression for the str annotation accepting a plain
```
eduardo-rodrigues:
```suggestion
    # 'pi' should not make 'pi0'/'pi+' stable.
```

tests/decay/test_decay.py
```diff
@@ -431,6 +450,33 @@ def test_DecayChain_flatten_with_stable_particles():
...
+    dm3 = DecayMode(0.98823, "gamma gamma")
+    dc = DecayChain("D0", {"D0": dm1, "K_S0": dm2, "pi0": dm3})
+    dc_flatten = dc.flatten(stable_particles=["D0", "pi0"])
+    assert dc_flatten.mother == "D0"
+    # K_S0 is still decayed, pi0 stays stable, D0 (mother) is decayed too.
+    assert dc_flatten.decays["D0"].daughters == DaughtersDict(["pi+", "pi-", "pi0"])
+
+
+def test_DecayChain_flatten_stable_particles_no_substring_match():
+    # stable_particles must not do substring matching: 'pi' should not make
```
eduardo-rodrigues:
```suggestion
    # stable_particles must not do substring matching:
```

tests/decay/test_decay.py
```diff
@@ -431,6 +450,33 @@ def test_DecayChain_flatten_with_stable_particles():
...
+    # 'pi0'/'pi+' stable (regression for the str annotation accepting a plain
+    # string and falling into substring containment).
```
eduardo-rodrigues:
```suggestion
```

eduardo-rodrigues:
Hi @henryiii, this brings in some important fixes and/or improvements. Thank you.

PR #585, opened 2026-06-11
chore: fix packaging config, CI, and pre-commit housekeeping

henryiii:

Part of #581.

## Summary

Ten targeted fixes to pyproject.toml, noxfile.py, .pre-commit-config.yaml, and .github/workflows/ci.yml — no functional changes to library code.

- **pyproject.toml** — corrected over-escaped `--doctest-glob=\\*.rst` (was literal `\*.rst`, fnmatch never matches it); removed dead pandas/PyArrow `filterwarnings` ignore (only existed in pandas 2.2.0/2.2.1, lockfile has 2.3.3+); aligned `[tool.pylint] py-version` with `requires-python` (`"3.10"` instead of `"3.12"`); raised `numpy>=1.21.6` and `pandas>=1.4` to the oldest versions that ship cp310 wheels on PyPI (verified against pypi.org/pypi/\*/json).
- **noxfile.py** — replaced `str(*Path("dist").glob("*.whl"))` (silent empty string / TypeError with multiple wheels) with `"dist/"` — `check-wheel-contents` accepts a directory directly. The now-unused `Path` import was auto-removed by ruff.
- **.pre-commit-config.yaml** — reduced mypy hook's `additional_dependencies` from `[attrs, particle, importlib_resources, numpy]` to `[particle, hepunits]` (only packages directly imported in `dec/`, `decay/`, `utils/` covered by the hook); removed the dead `requirements-txt-fixer` hook (no `requirements*.txt` files exist).
- **.github/workflows/ci.yml** — replaced maintenance-mode `pre-commit/action@v3.0.1` with `j178/prek-action@v2` (project already standardised on prek per noxfile and CLAUDE.md); added `fetch-depth: 0` to the `checks` job so hatch-vcs stamps the same version as the other jobs; folded the separate `python -m pip install pytest-github-actions-annotate-failures` step into the main `uv pip install` line (removes the mislabelled "Pytest" step that only installed, never tested).

## Validation

- `uv run pytest`: 299 passed, 1 skipped (no regressions from filterwarnings removal or glob fix)
- `prek -a --quiet`: clean (ruff auto-removed unused `Path` import in noxfile.py)
- `pylint src/`: 10.00/10 (py-version 3.10 is clean)
- `actionlint` / `zizmor`: actionlint not available via uvx; zizmor findings are all pre-existing unpinned-action warnings unrelated to this PR

## Open question: `decaylanguage[modeling]` extra

The `modeling/` sub-package pulls in `pandas`, `numpy`, and `plumbum`, which are currently unconditional runtime dependencies even for users who only need `.dec` parsing or the universal decay-chain representation. Splitting them into an optional `decaylanguage[modeling]` extra would reduce install weight for most users. This is a maintainer decision (it would be a breaking change for anyone not opting in) — flagging here for discussion, not implementing.

## Test plan

- [ ] CI passes on all matrix entries
- [ ] `nox -s build` produces a valid wheel and check-wheel-contents is satisfied
- [ ] Confirm prek-action runs correctly in the lint job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,9 +21,9 @@ jobs:
       with:
         python-version: "3.x"
     - uses: astral-sh/setup-uv@v8.1.0
-    - uses: pre-commit/action@v3.0.1
+    - uses: j178/prek-action@v2
       with:
-        extra_args: --all-files --hook-stage manual
+        extra-args: --all-files --hook-stage manual
     - name: PyLint
       run: uvx nox -s pylint -- --output-format=github
 
@@ -44,6 +44,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v6
+      with:
+        fetch-depth: 0
 
     - name: Setup Python ${{ matrix.python-version }}
       uses: actions/setup-python@v6
@@ -54,11 +56,7 @@ jobs:
     - uses: astral-sh/setup-uv@v8.1.0
 
     - name: Install package
-      run: uv pip install --system -e. --group=test
-
-    - name: Pytest
-      if: runner.os == 'Linux'
-      run: python -m pip install pytest-github-actions-annotate-failures
+      run: uv pip install --system -e. --group=test pytest-github-actions-annotate-failures
 
     - name: Test package
       run: python -m pytest ./tests --cov=src/decaylanguage --cov-report=xml
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,8 +44,8 @@ dependencies = [
     "attrs>=19.2",
     "graphviz>=0.12.0",
     "lark>=1.0.0",
-    "numpy>=1.9.3",
-    "pandas>=1",
+    "numpy>=1.21.6",
+    "pandas>=1.4",
     "particle>=0.26.0",
     "hepunits>=2.4.0",
     "plumbum>=1.7.0",
@@ -105,7 +105,7 @@ testpaths = ["tests"]
 addopts = [
     "-ra",
     "--doctest-modules",
-    "--doctest-glob=\\*.rst",
+    "--doctest-glob=*.rst",
     "--strict-markers",
     "--strict-config",
 ]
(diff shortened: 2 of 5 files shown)


henryiii:
`decaylanguage[modeling]` extra does sound attractive, it's something most users aren't using, I expect.

.github/workflows/ci.yml
```diff
@@ -21,9 +21,9 @@ jobs:
       with:
         python-version: "3.x"
     - uses: astral-sh/setup-uv@v8.1.0
-    - uses: pre-commit/action@v3.0.1
+    - uses: j178/prek-action@v2
```
eduardo-rodrigues, line 24: Does this change make sense, @henryiii? I would be inclined to stay with pre-commit.

pyproject.toml
```diff
@@ -44,8 +44,8 @@ dependencies = [
     "attrs>=19.2",
     "graphviz>=0.12.0",
     "lark>=1.0.0",
-    "numpy>=1.9.3",
-    "pandas>=1",
+    "numpy>=1.21.6",
+    "pandas>=1.4",
```
eduardo-rodrigues, line 48: One could even move to Pandas 2 at this point, but fine by now and we make the move after all your enhancements make it to a minor release. OK?

eduardo-rodrigues (approved):
Hi @henryiii, I left the odd little comment. Otherwise all good and happy if you merge.

eduardo-rodrigues:
> `decaylanguage[modeling]` extra does sound attractive, it's something most users aren't using, I expect.

Yep.

.github/workflows/ci.yml
henryiii: Yes, prek is fast, used on lots of projects like CPython, and the action is fully supported (pre-commit's action is in maintenance mode)

pyproject.toml
henryiii: Sounds good for a followup.

PR #586, opened 2026-06-11
refactor(modeling): attrs modernization and dead-code removal

henryiii:

## Summary

Scoped modernization and dead-code removal in the `modeling/` subsystem. Part of #581.

**Note:** A parallel modeling correctness-fix PR exists. Merge that one first; this PR rebases easily since it only touches style/structure, not logic.

- **attrs modernization**: `ModelDecay` (`decay.py`) and `AmplitudeChain` (`amplitudechain.py`) migrated from deprecated `@attr.s`/`attr.ib` to `@attrs.define`/`attrs.field`. `order=True` is passed explicitly to `attrs.define` to preserve the comparison/ordering semantics of the old `@attr.s` default (which defaulted to `order=True`; `attrs.define` defaults to `order=False`). All validators and `eq=False`/`order=False` per-field flags are preserved.
- **Dead graphviz guard removed**: `decay.py` had a `try/except ImportError` around `import graphviz` that emitted a `warnings.warn` (which would be an error under the test config anyway) and conditionally defined `_make_graphviz` and `_repr_mimebundle_` methods. Since `graphviz>=0.12.0` is a hard dependency in `pyproject.toml`, the guard is dead; both methods are now defined unconditionally.
- **Unused locals dropped**: `make_lineshape` in both `GooFitChain` and `GooFitPyChain` computed `a` and `b` from `structure[0]+1` / `structure[1]+1`, swapped them if out of order, then never used them. Removed.
- **Unreachable branches removed**: `make_spinfactor` in both classes had an `if not spin_factors: …` guard, but `self.spinfactors` raises `LineFailure` rather than returning an empty list when the spin factor is unknown — so the guard can never be reached. Removed, leaving the `for spin_factor in spin_factors` loop unconditional.
- **`elif` in `__main__.py`**: The two sequential `if` checks on `self.generator` are mutually exclusive (it's a `cli.Set("goofit", "goofitpy")`) — changed second `if` to `elif`.

## Test plan

- [x] `uv run pytest` — 299 passed, 1 skipped (no changes to generated output in `tests/output/`)
- [x] `prek -a --quiet` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* Modeling subsystem (`modeling/`):
+  - Migrated `ModelDecay` and `AmplitudeChain` from old-style `@attr.s`/`attr.ib` to modern `@attrs.define`/`attrs.field`.
+  - Removed dead `graphviz` import guard in `decay.py` (graphviz is a hard dependency).
+  - Removed unused local variables `a`, `b` in `make_lineshape` methods of `GooFitChain` and `GooFitPyChain`.
+  - Removed unreachable `if not spin_factors:` branches in `make_spinfactor` (`spinfactors` raises `LineFailure` rather than returning empty).
+  - Fixed sequential `if` checks in `__main__.py` `main` to use `elif`.
+
 * CI and tests:
   - Several improvements, enhancements and clean_ups.
   - Removed dead `filterwarnings` ignore for PyArrow/pandas deprecation (pandas >=2.2.2 no longer emits it).
--- a/src/decaylanguage/__main__.py
+++ b/src/decaylanguage/__main__.py
@@ -21,7 +21,7 @@ class DecayLanguageDecay(cli.Application):
     def main(self, filename):
         if self.generator == "goofit":
             ampgen2goofit(filename)
-        if self.generator == "goofitpy":
+        elif self.generator == "goofitpy":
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+* Modeling subsystem (`modeling/`):
+  - Migrated `ModelDecay` and `AmplitudeChain` from old-style `@attr.s`/`attr.ib` to modern `@attrs.define`/`attrs.field`.
+  - Removed dead `graphviz` import guard in `decay.py` (graphviz is a hard dependency).
+  - Removed unused local variables `a`, `b` in `make_lineshape` methods of `GooFitChain` and `GooFitPyChain`.
+  - Removed unreachable `if not spin_factors:` branches in `make_spinfactor` (`spinfactors` raises `LineFailure` rather than returning empty).
+  - Fixed sequential `if` checks in `__main__.py` `main` to use `elif`.
```
henryiii: These changelog entries are too detailed (and they will all collide), worth trimming down as these get merged.

eduardo-rodrigues: Indeed. I simplified a bit.

eduardo-rodrigues (approved):
Hi @henryiii, this looks good to me but do check the bits removed in modeling. Am already approving; merge when/if happy. Thank you.

PR #588, opened 2026-06-11
refactor(dec): typing modernisations and simplifications in dec/ subsystem

henryiii:

Part of #581

## Summary

Mechanical simplifications and typing modernisations in the `dec/` subsystem, touching only `src/decaylanguage/dec/dec.py` and `src/decaylanguage/dec/enums.py`:

- **Typing fixes** (`dec.py`):
  - `stable_particles` annotation on `build_decay_chains`: `list[str] | set[str] | tuple[str] | tuple[()]` → `Collection[str]` (imported from `collections.abc`).
  - `_align_items` parameter: `str` → `list[str] | list[list[str]]` to match actual call sites; added a narrowing local variable so mypy can type-check the element operations.
  - `_grammar_info`, `_parsed_decays`: normalised `None | (dict[str, Any])` / `None | (Any)` → `dict[str, Any] | None` / `Any`.
  - `list_decay_mother_names` return: `list[str | Any]` → `list[str]`.
  - `get_decay_mother_name` return: `str | Any` → `str` (with `cast()`).
  - `_include_ccdecays`: removed spurious `or False` (parameter is already `bool`).
  - `DecayModelParamValueReplacement` / `ChargeConjugateReplacement`: parameterised as `Visitor[Token]`; `# type: ignore[misc]` retained because lark's `Visitor` is re-exported as `Any`.
  - `DecayModelAliasReplacement`: kept as `Transformer` with `# type: ignore[misc]` (parameterisation introduced new errors in the untouchable body).

- **Collapsed four near-identical two-token extractors** into a single private helper `_extract_two_token_dict(rule, *, value_converter, nested_tokens)`. Public signatures, docstrings, and any doctests are unchanged.

- **`setdefault` simplifications**:
  - `get_pythia_definitions`: replaced if/else-update pattern.
  - `get_jetset_definitions`: replaced try/except-`KeyError` insertion pattern.

- **`enums.py` comment** updated to describe the `edit_terminals` injection mechanism rather than the now-stale grammar-sync note.

## Test plan

- [x] `uv run pytest` — 299 passed, 1 skipped (pre-existing unconditional skip)
- [x] `prek -a --quiet` — all hooks pass (ruff format + mypy)
- [x] mypy error count: 100 before → 92 after (net improvement; no new errors introduced)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+* Parsing of decay files (aka .dec files):
+  - Typing modernisations in `dec/dec.py`: `Collection[str]` for `stable_particles`,
+    `list[str] | list[list[str]]` for `_align_items`, proper generics on `Visitor[Token]`
+    subclasses, and normalised `None | (X)` constructs.
+  - Collapsed four near-identical two-token extractors (`get_definitions`,
+    `get_aliases`, `get_charge_conjugate_defs`, `get_decays2copy_statements`)
+    into a shared private helper `_extract_two_token_dict`.
+  - Simplified `get_pythia_definitions` and `get_jetset_definitions` to use
+    `dict.setdefault` instead of if/else-update and try/except-KeyError patterns.
+  - Updated the `known_decay_models` comment in `dec/enums.py` to reflect that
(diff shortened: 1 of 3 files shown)


eduardo-rodrigues (approved):
Various very nice improvements, thank you.

PR #590, opened 2026-06-11
fix(modeling): correct AmpGen to GooFit converter bugs

henryiii:

Part of #581.

Fixes a cluster of correctness bugs in the `modeling/` subsystem (the AmpGen to GooFit converter) and regenerates the baked reference output `tests/output/DtoKpipipi_v2.cu`.

> [!WARNING]
> The generated GooFit output changes **semantically**: amplitude fix/free flags are flipped where AmpGen marked a coefficient `Fix=2`, propagated coefficient errors change, and the S-wave `A -> V P` spin factor changes. Downstream users who generated fitters from these conversions should regenerate and re-check their fits.

## Per-fix summary

1. **Amplitude fix-flag inversion** (`ampgentransform.py`): `cplx_decay_line` computed `decay["fix"] = not (real_free and imag_free)` against values that are actually *is-fixed* booleans, so the flag was inverted. Now `decay["fix"] = bool(real_fixed and imag_fixed)`, consistent with the C++ `make_pars` path (`if not par.fix`). Removed the dead, inverted `fixed`/`free` transformer methods (the grammar only ever produces `checkfixed`).
2. **Polar to cartesian error propagation** (`amplitudechain.py`): replaced `sin(dtheta)`/`cos(dtheta)` with `sin(theta)*dtheta`/`cos(theta)*dtheta`. Sanity-checked: `A=1, dA=0, dtheta=0` now yields `0+0j` (previously `~1j`).
3. **`GooFitPyChain.make_amplitude`** (`goofit.py`): the imaginary free coefficient is now named `_i` (was `_r`), and cartesian components use unbounded `Variable(name, value, error)` instead of the `0., 1000.` bounds (cartesian components are frequently negative), mirroring the C++ `mkvar` semantics.
4. **`ampgen2goofitpy` lost output with `ret_output=True`** (`ampgen2goofit.py`): bare `print` calls for the header and the final `DK3P_DI.amplitudes = amplitudes_list` line now go through `printer`.
5. **`FastCoherentSum::UseCartesian` parser crash** (`amplitudechain.py`): fixed the tuple unpacking (`((fcs,),) = fcs`) and use `bool(int(fcs))` so `UseCartesian 0` is correctly falsy. Added regression tests for `0` and `1`.
6. **Particle CSV reloaded for every node** (`amplitudechain.py`): the old guard compared an int never present in the CSV against `Particle` objects, reloading the table on every recursive call. Now guarded with `Particle.from_pdgid(998101)`, using `data.basepath` for the path; dropped the `particle` 0.4.4 compat shim and the unused `os` import.
7. **Shared mutable class state** (`amplitudechain.py`): `read_ampgen` now resets `all_particles`, `final_particles`, and `cartesian` on `cls` at the start, so repeated calls and the two sibling subclasses (`GooFitChain`/`GooFitPyChain`) no longer leak stale particles or a stale Cartesian flag. Added a regression test that parsing the same input twice yields identical results.
8. **`L_range` missed intermediate couplings** (`amplitudechain.py`): now computes `min L = 0` when `|s1-s2| <= S <= s1+s2`, else `min(|S-|s1-s2||, |S-(s1+s2)|)`; `max L = S+s1+s2`. Added a unit test (spin-1 to two vectors gives `(0, 3)`).
9. **S-wave A to VP mapped to D-wave spin factor** (`goofit.py`): `"DtoA1P1_A1toV2P2_V2toP3P4"` now maps to the S-wave enum `SF_4Body.DtoAP1_AtoVP2_VtoP3P4` instead of the D-wave one (the D-wave key on the line above is unchanged).
10. **ANSI color leak** (`ampgen2goofit.py`): a `paint()` helper only applies color styling when writing to a real stdout; with `ret_output=True` no escape codes reach the generated file.

## Reference-output diff analysis (`tests/output/DtoKpipipi_v2.cu`)

The test compares the regenerated output to the reference as a **set of stripped lines** (ignoring the `Generated on` timestamp), so block reordering and trailing whitespace are not significant. Every change in the regenerated file falls into one of these expected categories, with nothing unexplained:

- **Fix flags flipped** (66 `mkvar` lines): coefficients AmpGen marked `2` (Fix) now emit `true`, and those marked `0` (Free) now emit `false` -- verified line-by-line against `models/DtoKpipipi_v2.txt`.
- **Corrected coefficient errors** (same `mkvar` lines): fixed coefficients with `dA=0, dtheta=0` now show imaginary error `0.0` (was `1.0`); free coefficients show the corrected first-order propagated errors.
- **S-wave spin factor** (the S-wave `A -> V P` lines, e.g. `K(1)(1270)-[GSpline.EFF]{...}` and `K(1)(1400)-{...}`): now emit `SF_4Body::DtoAP1_AtoVP2_VtoP3P4`; the explicit D-wave lines (`[D;GSpline.EFF]`) correctly keep `DtoAP1_AtoVP2Dwave_VtoP3P4`.
- **`L = 0` formatting**: `L_range` now returns an integer `0` for the in-window minimum, so `L: 0.0`/`[0.0-...]` and lineshape arguments like `..., 0.0, M_24, ...` now print `0` instead of `0.0`. Cosmetic only.
- **Block reordering / trailing whitespace**: spin-configuration listing and the spline-array blocks (iterated from sets) appear in a different order; content is identical.

## Testing

- `uv run pytest` -- 305 passed, 1 skipped.
- `prek -a --quiet` -- clean.


eduardo-rodrigues:
Hi @henryiii, I'm done with my reviews as I leave this one to you, as you're the developer of this submodule.

We are reaching the point for a major 1.0 release. We keep in touch for this :).

PR #594, opened 2026-06-16
refactor(typing): reduce type: ignore comments (#435)

### LINKED ISSUE #435: Reduce or remove completely type: ignore statements
Indeed they could potentially hide real issues with the code. For a start all type: ignore should be reviewed.

henryiii:
Addressing an old issue.

Closes #435.

Reviewed every `# type: ignore` in the mypy-checked subsystems (`dec/`, `decay/`, `utils/`) and removed **21 of 34**, replacing each with a proper annotation or a real runtime guard instead of a silenced error.

## Changes

**`dec/dec.py` (15 removed)**
- `_dec_file` retyped to `str | None` (declared once before the `if`), dropping the `[assignment]` ignore.
- `_parsed_decays` retyped from `Any | None` → `list[Tree] | None`. The 12 `union-attr` / `arg-type` / `index` ignores across `parse`, `_add_decays_to_be_copied`, `_add_charge_conjugate_decays`, `_check_parsed_decays`, `number_of_decays`, `list_decay_mother_names`, and `_find_decay_modes` are replaced by `assert self._parsed_decays is not None`. This is the heart of the issue's concern — those ignores silently hid the `None` (called-before-parse) case; the asserts turn it into a clear failure.
- `grammar()` now asserts `self._grammar is not None`, matching the existing pattern in the neighbouring `grammar_info()`.

**`decay/decay.py` (1 removed)**
- `DaughtersDict.fromkeys` annotated as `(cls, iterable: Any, v: Any = None) -> NoReturn`.

**`decay/viewer.py` (5 removed)**
- `to_string()` returns `str(self.graph.source)`.
- The 4 `[arg-type]` ignores in `iterate_chain` are gone after asserting the chain-dict value types.

## Deliberately kept (13)

Inherent to untyped third-party code or intentional behaviour, not maskable real bugs:
- `decay.py` — `[override]` on `DaughtersDict.__add__` / `__sub__` (narrowing `other` to `Self` is intentional and incompatible with `Counter`'s signature).
- `dec.py` — 3× `[misc]` on subclassing lark's untyped `Transformer` / `Visitor`.
- `dec.py` — `[operator]` (particle width) and `[union-attr]` (jetset regex match) — both inside `try/except` that intentionally handles the `None` case (the latter documented "will throw an error if match is None").
- TypedDict / `Sequence` narrowing spots (`decay.py:422,747,984`, `dec.py:1050,1219`, `viewer.py:256`) where a fix would require widening shared type definitions with cascading risk.

## Verification
- `prek -a --quiet` clean (incl. mypy via the CI hook config, where lark is untyped)
- `pytest`: 332 passed, 1 skipped

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Typing:
+  - Reviewed and reduced `# type: ignore` comments across `dec/`, `decay/`, and
+    `utils/` (#435). Several were replaced by proper annotations or `assert`
+    invariants (e.g. on `DecFileParser._parsed_decays`), turning silently
+    ignored errors into real runtime guards.
 * Parsing of decay files (aka .dec files):
   - Various improvements to the code, for more robustness.
   - Performance improvements in `DecFileParser`, with caching and lazily-built indexing where possible.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,6 @@
 
 ## Unreleased
 
-* Typing:
-  - Reviewed and reduced `# type: ignore` comments across `dec/`, `decay/`, and
-    `utils/` (#435). Several were replaced by proper annotations or `assert`
-    invariants (e.g. on `DecFileParser._parsed_decays`), turning silently
-    ignored errors into real runtime guards.
 * Parsing of decay files (aka .dec files):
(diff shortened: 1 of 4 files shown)


eduardo-rodrigues (approved):
This is great to see. It's been on the to-dos list for a while, at low(er) priority. Thank you very much.

PR #602, opened 2026-06-29
Better coverage of decay file parsing helper functions

eduardo-rodrigues:


eduardo-rodrigues:
Hey @henryiii, I noticed that the coverage report has not been posted for a while here while in Particle it's been working fine. I did notice the misconfiguration from past time when we were working on the master rather than main branch. I fixed this before creating this PR but the coverage report still doesn't appear here. Do you have any ideas on what to check further?

eduardo-rodrigues:
I added this package explicitly to the list of repos with Codecov access. Funny enough Particle was also not there and I added it too. Fingers crossed it's all fine in both repos!

PR #606, opened 2026-07-06
Little update of dependencies vis-a-vis Particle

### LINKED ISSUE #600: Update dependencies to sync better with Particle
Hi @henryiii, I noticed that the dependencies in particular here in the `modeling` submodule,
```
modeling = [
    "numpy>=1.21.6",
    "pandas>=2.0",
    "plumbum>=1.7.0",
]
```
and in relation to the dependencies in Particle, aren't quite in sync as they could be. We should tune the dependencies a little bit. Let me know if you agree, especially in relation to the `modeling` submodule.

eduardo-rodrigues:
Closes https://github.com/scikit-hep/decaylanguage/issues/600.

This is pretty small an update, and I think it makes sense - makes the search matrix smaller. But see if you @henryiii have a second to cross-check.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,9 +5,9 @@ dependencies:
   - attrs>=22.2.0
   - pip>=9
   - jupyterlab>=0.4
-  - numpy>=1.19.3
+  - numpy>=1.26.0
   - graphviz>=0.12.0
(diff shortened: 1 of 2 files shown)


eduardo-rodrigues:
Going forward, merging.

PR #607, opened 2026-07-06
Update development status to Production/Stable

eduardo-rodrigues:
This was due since we release 1.0.0 :).


eduardo-rodrigues (commented):
Merging since a trivial update.

